PR #1, opened 2025-05-23
dependencies

loftusa:
faiss/tqdm


norabelrose:
hey thanks, will add the tqdm dependency shortly, but we don't want to depend on the PyPI faiss-cpu package because it's unofficial and may not be updated in the future

PR #8, opened 2025-06-19
Batch-based processing filter_data.py that calculates mean gradients

dipikakhullar:
1. Batch-based processing system that calculates mean gradients, normalizes gradients, and computes inner products in smaller batches…zes gradients, and computes inner products in smaller batches

### DIFF
--- a/examples/filter_data.py
+++ b/examples/filter_data.py
@@ -13,8 +13,9 @@ from datasets import load_dataset, Dataset
 from transformers import AutoTokenizer, AutoModelForCausalLM
 from trl import SFTConfig, SFTTrainer, setup_chat_format
 
-from bergson.data import load_index, tokenize
+from bergson.data import load_gradient_dataset, tokenize, DataConfig
 from bergson.utils import assert_type
+import numpy as np
 
 
 @dataclass
@@ -81,7 +82,7 @@ def select_topk(ds: Dataset, n: int, key: str, lowest: bool = False):
 
 def normalize_batch(batch):
     return {
-        "gradient": F.normalize(batch["gradient"].cuda(), dim=1).cpu(),
+        "gradients": F.normalize(batch["gradients"].cuda(), dim=1).cpu(),
     }
 
 
@@ -92,13 +93,9 @@ def add_index(
     assert index is not None, "Index must be provided for attribution or loss filtering"
 
     gradient_ds = (
-        load_index(index)
+        load_gradient_dataset(index)
         .with_format("torch")
-        .map(normalize_batch, batched=True, batch_size=512)
-        .sort("row_number")
-        .select_columns(["gradient", "loss", "row_number"])
     )
-
     def generator():
         for row, grad_row in zip(dataset, gradient_ds):
             yield {**dict(row), **dict(grad_row)}
@@ -125,15 +122,33 @@ def main(
 
     with nullcontext() if rank == 0 else redirect_stdout(None):
         if args.query_index and args.filter == "attribution":
-            print("loading index...")
-            query_index = (
-                load_index(args.query_index)
-                    .with_format("torch")
-                    .map(normalize_batch, batched=True, batch_size=512)      
-            )
+            print("loading query index...")
+            query_index = load_gradient_dataset(args.query_index).with_format("torch")
+            
+            # Calculate mean gradient in batches to avoid memory issues
+            print("Computing query mean gradient...")
+            mean_gradient = None
+            batch_size = 512
+            
+            for i in tqdm(range(0, len(query_index), batch_size), desc="Computing query mean"):
+                batch = query_index.select(range(i, min(i + batch_size, len(query_index))))
+                batch_gradients = torch.stack([g for g in batch["gradients"]])
+                
+                # Normalize gradients
+                batch_gradients = F.normalize(batch_gradients, dim=1)
+                
+                if mean_gradient is None:
+                    mean_gradient = batch_gradients.mean(0)
+                else:
+                    mean_gradient = (mean_gradient * i + batch_gradients.sum(0)) / (i + len(batch))
+                
+                # Clear memory
+                del batch_gradients
+            
+            query = mean_gradient.cuda()
+            del query_index, mean_gradient
+            torch.cuda.empty_cache()
             print("query index loaded")
-            query = assert_type(Tensor, query_index["gradient"]).mean(0).cuda()
-            del query_index
         else:
             query = None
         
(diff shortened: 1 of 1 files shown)


examples/filter_data.py
```diff
@@ -125,15 +122,33 @@ def main(
...
+                    mean_gradient = batch_gradients.mean(0)
+                else:
+                    mean_gradient = (mean_gradient * i + batch_gradients.sum(0)) / (i + len(batch))
+                
+                # Clear memory
+                del batch_gradients
+            
+            query = mean_gradient.cuda()
+            del query_index, mean_gradient
+            torch.cuda.empty_cache()
```
luciaquirke, line 150: Many people in the alignment community add the `torch.cuda.empty_cache()` line when they're having memory leaks they don't know how to fix, but it doesn't fix the leaks or achieve anything in most contexts. They are not setting a good example, plz do not adopt 😅 

My understanding is that `torch.cuda.empty_cache()` releases VRAM reserved by PyTorch that it doesn't need anymore, so if you are simultaneously running other processes that need VRAM it can be helpful.

examples/filter_data.py
```diff
@@ -191,30 +246,42 @@ def add_rank(ex):
...
             raise ValueError(f"Invalid filter: {args.filter}")
 
+        # Clear memory before loading model
+        torch.cuda.empty_cache()
+        
+        print("Loading model...")
         model = AutoModelForCausalLM.from_pretrained(
             args.model,
             torch_dtype="bfloat16",
+            low_cpu_mem_usage=True,
```
luciaquirke, line 264: Does this do anything? Seems deprecated

examples/filter_data.py
```diff
@@ -81,7 +82,7 @@ def select_topk(ds: Dataset, n: int, key: str, lowest: bool = False):
 
 def normalize_batch(batch):
```
luciaquirke, line 83: Can be removed

examples/filter_data.py
```diff
@@ -125,15 +122,33 @@ def main(
...
+                batch = query_index.select(range(i, min(i + batch_size, len(query_index))))
+                batch_gradients = torch.stack([g for g in batch["gradients"]])
+                
+                # Normalize gradients
+                batch_gradients = F.normalize(batch_gradients, dim=1)
+                
+                if mean_gradient is None:
+                    mean_gradient = batch_gradients.mean(0)
+                else:
+                    mean_gradient = (mean_gradient * i + batch_gradients.sum(0)) / (i + len(batch))
```
luciaquirke, line 143: Mean calculation is incorrect

examples/filter_data.py
```diff
@@ -125,15 +122,33 @@ def main(
...
+            print("Computing query mean gradient...")
+            mean_gradient = None
+            batch_size = 512
+            
+            for i in tqdm(range(0, len(query_index), batch_size), desc="Computing query mean"):
+                batch = query_index.select(range(i, min(i + batch_size, len(query_index))))
+                batch_gradients = torch.stack([g for g in batch["gradients"]])
+                
+                # Normalize gradients
+                batch_gradients = F.normalize(batch_gradients, dim=1)
```
luciaquirke, line 138: To get performance benefits from this method we need to move each batch to CUDA

examples/filter_data.py
```diff
@@ -160,20 +175,60 @@ def main(
             pass
         elif args.filter == "attribution":
             if query == None:
-                query = train["gradient"].mean(0).cuda()
+                # Calculate mean gradient in batches to avoid memory issues
+                print("Calculating mean gradient...")
+                mean_gradient = None
+                batch_size = 100  # Process in smaller batches
```
luciaquirke, line 181: 512, I reckon also add the batch size to FilterConfig like we talked about

examples/filter_data.py
```diff
@@ -160,20 +175,60 @@ def main(
...
-                Dataset, Dataset.from_generator(importance_score_generator)
-            )
-            train = select_topk(train, args.num_examples, "importance_score", lowest=args.lowest)
+            print("Computing importance scores...")
+            # Calculate importance scores in batches
+            importance_scores = []
+            batch_size = 512
+            
+            for i in tqdm(range(0, len(train), batch_size), desc="Computing importance scores"):
+                batch = train.select(range(i, min(i + batch_size, len(train))))
```
luciaquirke, line 203: I _think_ you can do `train[i : min(i + batch_size, len(train))]` then remove the `torch.stack`

examples/filter_data.py
```diff
@@ -160,20 +175,60 @@ def main(
...
-                        "importance_score": (row["gradient"].cuda() @ query).item(),
-                    }
-
-            train = assert_type(
-                Dataset, Dataset.from_generator(importance_score_generator)
-            )
-            train = select_topk(train, args.num_examples, "importance_score", lowest=args.lowest)
+            print("Computing importance scores...")
+            # Calculate importance scores in batches
+            importance_scores = []
```
luciaquirke, line 199: Preallocate torch.zeros vector of len(train)

examples/filter_data.py
```diff
@@ -160,20 +175,60 @@ def main(
...
+            for i in tqdm(range(0, len(train), batch_size), desc="Computing importance scores"):
+                batch = train.select(range(i, min(i + batch_size, len(train))))
+                batch_gradients = torch.stack([g for g in batch["gradients"]]).cuda()
+                
+                # Normalize gradients
+                batch_gradients = F.normalize(batch_gradients, dim=1)
+                
+                # Compute inner products
+                batch_scores = (batch_gradients @ query).cpu().numpy()
+                importance_scores.extend(batch_scores)
```
luciaquirke, line 211: Add to tensor using the same indexing as for the batch selection

examples/filter_data.py
```diff
@@ -191,30 +246,42 @@ def add_rank(ex):
...
+        print("Loading model...")
         model = AutoModelForCausalLM.from_pretrained(
             args.model,
             torch_dtype="bfloat16",
+            low_cpu_mem_usage=True,
         )
         tokenizer = AutoTokenizer.from_pretrained(args.model, max_length=8192)
         model, tokenizer = setup_chat_format(model, tokenizer)
 
+        # Create DataConfig for tokenization
```
luciaquirke, line 269: Nice

examples/filter_data.py
luciaquirke: If you don't have cursor autocomplete enabled I recommend enabling it cause it should be able to help with this

PR #10, opened 2025-06-19
Small fixes

LouisYRYJ:
- Fix int8 and FSDP compatiblity
- Use memmap mode=r+ instead of w+ to prevent racing issues
- Fix gradient processor loading/saving logic


bergson/build.py
```diff
@@ -163,7 +177,7 @@ def build_index(cfg: IndexConfig):
                 ds = Dataset.load_from_disk(data_str, keep_in_memory=False)
             else:
                 raise e
-
+    # ds = ds.select(list(range(73)))
```
LouisYRYJ, line 180:
```suggestion
```

PR #13, opened 2025-06-22
Ekfac

LouisYRYJ:
This introduces EKFAC computations, everything relevant is in /hessians (ignore /approx_unrolling for now)
I mostly followed the build_index pipeline.
Currently, we should not merge this, because it uses the current version main and not the ultrafast one. 
As I refactored some parts of the general pipeline, we should probably merge ultrafast first and then I can make the changes to this PR.


bergson/distributed.py
```diff
@@ -0,0 +1,96 @@
...
+    ds = ds.map(
+        tokenize,
+        batched=True,
+        fn_kwargs=dict(args=cfg.data, tokenizer=tokenizer),
+        remove_columns=remove_columns,
+    )
+
+    world_size = torch.cuda.device_count() if cfg.world_size is None else cfg.world_size
+    if world_size <= 0:
+        warning_msg = f"world_size is set to {world_size}, but must be positive. Setting world_size=1."
```
norabelrose, line 61: This seems a little weird, since it means that if `world_size` is set to 1 on the command line then it will use distributed stuff, but if you set it to 0 then it won't. Is that the expected behavior?

bergson/hessians/collector.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Before we add any hooks, we need to peek at what modules we need to track.
+        for name, layer in self.model.named_modules():
+            if not isinstance(layer, nn.Linear):
+                continue
+
+            if self.target_modules is not None and name not in self.target_modules:
+                continue
+
+            if "mlp" not in name:
```
norabelrose, line 61: This is fairly fragile but I'm not sure how else to handle it with HuggingFace

bergson/hessians/compute_EKFAC.py
```diff
@@ -0,0 +1,202 @@
...
+        buf, chunk_id = [], 0
+
+        def flush():
+            nonlocal buf, chunk_id
+            if not buf:
+                return
+            sub_ds = assert_type(Dataset, Dataset.from_list(buf))
+            batches = allocate_batches(sub_ds["length"], cfg.token_batch_size)
+
+            compute_all_factors(
```
norabelrose, line 148: hmm it seems like this file is almost entirely copied from `build.py`, so maybe we want to refactor to reduce code redundancy? In particular instantiating the model, init process group, the PEFT stuff, etc. can be shared I think

bergson/distributed.py
LouisYRYJ: thanks for catching this, will fix

bergson/hessians/compute_EKFAC.py
LouisYRYJ: Yeah good idea, will do that

PR #14, opened 2025-06-24
Support streaming with IterableDataset

luciaquirke:
- Support IterableDataset streaming
   - chunk the IterableDataset into Dataset objects of configurable size for gradient collection
 - Add config option to specify model revision

Currently the IterableDataset is chunked for gradient collection but not for fitting normalizers, under the assumption that users will not use more data than can fit on disk for this step. If this doesn't work we can switch to using the gradient collection chunking method to fit normalizers too.

Trivia
 - Data filtering tweaks (automatic model upload, sampling with temperature)
 - Fix type of `stats_sample_size`
 - Add .vscode to .gitignore
 - Fix `drop_columns` so it does what it says on the tin (removes original columns, leaves the new column "length" untouched)

### DIFF
--- a/bergson/build.py
+++ b/bergson/build.py
@@ -6,7 +6,7 @@ from tqdm.auto import tqdm
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
-from datasets import Dataset, IterableDataset, load_dataset
+from datasets import Dataset, IterableDataset, DatasetDict, IterableDatasetDict, load_dataset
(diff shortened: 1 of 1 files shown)


bergson/build.py
```diff
@@ -153,6 +190,10 @@ def dist_worker(rank: int, world_size: int, cfg: IndexConfig, ds: Dataset):
         dist.destroy_process_group()
 
 
+def add_row(_, i):
+    return dict(_row=i)
```
luciaquirke, line 194: inline lambdas produce pickling errors

PR #20, opened 2025-07-18
Add FAISS

luciaquirke:
We can't write to a memmap larger than the CommitLimit on CoreWeave nodes (without adding swap), but once an equivalent number of rows have been written to across multiple memmaps we can load and concatenate them up to the RAM limit. 

Therefore at least initially we will support loading indices from multiple memmaps and then concatenating them in memory (works for ~1M rows). The larger indices (10M rows+) can be loaded from disk into a compressed-format GPU FAISS index in batches using the gradients_loader.

```
❯ grep ^Commit /proc/meminfo
CommitLimit:    395.637292 GB
Committed_AS:   26.360692 GB

❯ free -h
               total        used        free      shared  buff/cache   available
Mem:           754Gi       103Gi       406Gi       9.5Gi       243Gi       6
```

### DIFF
--- a/bergson/processing.py
+++ b/bergson/processing.py
@@ -41,6 +41,7 @@ def collect_gradients(
 
     # Mutable state for the GradientCollector callback
     mod_grads = {}
+    mod_grad_sum_sqs = {}
     preconditioners = {}
 
     # TODO: Handle this more elegantly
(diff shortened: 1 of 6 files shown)


bergson/processing.py
```diff
@@ -111,13 +110,19 @@ def callback(name: str, g: torch.Tensor):
 
             model.zero_grad()
 
+        norm = (
```
luciaquirke, line 113: we can save the norms if anyone wants to use them

pyproject.toml
```diff
@@ -25,6 +25,9 @@ dev = [
 example = [
     "trl",
 ]
+faiss = [
+    "faiss-gpu-cu12",
```
luciaquirke, line 29: can support CPU-only FAISS in the future (just need to set device to CPU or use the mmap config option)

PR #26, opened 2025-08-05
Add gradient collection trainer callback

luciaquirke:
Add gradient collection callback + minimal usage example

- Support multiple epochs, averaging or with aggregation disabled. Defaults to not aggregating because it's the non-destructive option but open to averaging as a sensible default?
- Support not normalizing gradients during validation set collection as in LESS
- Support Trainer and SFTTrainer

TODO:
- test with/support other Trainer subclasses

Follow up:
- test/scale to 10B+ token datasets


bergson/callback.py
```diff
@@ -0,0 +1,198 @@
...
+        self.mod_grads.clear()
+
+    def on_epoch_begin(
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
+        epoch = int(state.epoch or 0)
```
luciaquirke, line 74: Epoch is a float where the decimal part represents progress through the current epoch

PR #35, opened 2025-09-05
Add Dr. GRPO advantage estimates

luciaquirke:
Add utility flag to use the policy gradient loss when rewards are available


norabelrose (approved):
LGTM

PR #40, opened 2025-09-17
Enable per-head gradients

luciaquirke:
- Fix bug where projection device was not correct for multi-GPU training setups
- Enable per-head gradients in the callback and static build
- Add some documentation to the README

Notes
- Could add h.*.attn... formatting


norabelrose (approved):
LGTM thanks

PR #41, opened 2025-09-22
Fix ETA accuracy

luciaquirke:
Shuffling the batches empirically stabilizes the static index build ETA.

### DIFF
--- a/bergson/data.py
+++ b/bergson/data.py
@@ -1,6 +1,7 @@
 import json
 import math
 import os
+import random
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Literal, Sequence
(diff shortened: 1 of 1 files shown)


bergson/data.py
```diff
@@ -114,7 +115,7 @@ def ceildiv(a: int, b: int) -> int:
     return -(-a // b)  # Equivalent to math.ceil(a / b) but faster for integers
 
 
-def allocate_batches(doc_lengths: list[int], N: int) -> list[list[int]]:
+def allocate_batches(doc_lengths: list[int], N: int, seed: int = 42) -> list[list[int]]:
```
luciaquirke, line 118: Should we leave the seed to the end user?

PR #47, opened 2025-10-13
Add on-the-fly queries

### LINKED ISSUE #49: Use .part extension for in-progress index runs
Add logic to indicate when a build successfully completes so there's no chance of trying to using an incomplete index.

Firefox uses the .part extension for this then renames when it completes.

luciaquirke:
- Add on the fly queries
- Precompute query dataset if not already available
- Use .part extension of in-progress runs (closes #49)
- You can technically now torch.compile the model when projection_dim=0 and save_index=False but this slows down the build and projection_dim=0 is super memory hungry

TODO
- Extract out the query dataset assembly into an example script and consider making it a more official tool in the future

**Notes**

If keeping the extra gradients in VRAM before the query callback causes problems we can add something like

    offload_to_cpu: bool = False
    """If True, keep value gradients on CPU until the query callback is called."""

But it's only a small memory usage increase.

### DIFF
--- a/bergson/query.py
+++ b/bergson/query.py
@@ -38,8 +38,8 @@ from .utils import assert_type, get_layer_list
 
 def get_query_data(index_cfg: IndexConfig, query_cfg: QueryConfig):
     """
-    Assemble the preconditioners for the query dataset. Preconditioners
-    can be mixed as described in https://arxiv.org/html/2410.17413v1#S3.
+    Load and optionally precondition the query dataset. Preconditioners
+    may be mixed as described in https://arxiv.org/html/2410.17413v1#S3.
     """
     # Collect the query gradients if they don't exist
     if not os.path.exists(query_cfg.run_path):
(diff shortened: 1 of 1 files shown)


bergson/collection.py
```diff
@@ -53,9 +55,11 @@ def collect_gradients(
 
     def callback(name: str, g: torch.Tensor):
         g = g.flatten(1).clamp_(lo, hi)
-
-        # Asynchronously move the gradient to CPU and convert to fp16
-        mod_grads[name] = g.to(device="cpu", dtype=dtype, non_blocking=True)
+        if save_index:
```
luciaquirke: Avoid the round trip to cpu

bergson/data.py
```diff
@@ -111,7 +131,7 @@ class IndexConfig:
     """Whether to use Fully Sharded Data Parallel (FSDP) for collecing gradients."""
 
     precision: Literal["auto", "bf16", "fp16", "fp32", "int4", "int8"] = "auto"
-    """Precision to use for the model parameters."""
+    """Precision (dtype) to use for the model parameters."""
```
luciaquirke: improve searchability

bergson/collection.py
```diff
@@ -90,6 +94,12 @@ def callback(name: str, g: torch.Tensor):
         dtype=dtype,
         fill_value=0.0,
     )
+    per_doc_scores = torch.full(
```
luciaquirke: Only support one score per doc, i.e. don't support computing module scores separately for now

PR #50, opened 2025-10-16
add conv1d layer type (used in gpt2)

zywilliamli:
gpt2 type models use transformers.pytorch_utils.Conv1D instead of nn.Linear which is not currently tracked by gradient collector, this pr adds that layer type. 

tested manually using python -m bergson runs/test --model openai-community/gpt2 --dataset NeelNanda/pile-10k --truncation, also added unit tests for it.

### DIFF
--- a/bergson/gradients.py
+++ b/bergson/gradients.py
@@ -477,7 +477,7 @@ class GradientCollector(ContextDecorator):
     def _process_grad(self, module: nn.Module, _, grad_out):
         """Process the incoming gradient wrt the output of the module."""
         # Sanity checks
-        assert isinstance(module, LayerAdapter.supported_modules), "Expected a supported module"
+        assert isinstance(module, LayerAdapter.supported_modules), f"Expected a module of type {LayerAdapter.supported_modules}, got {type(module)}"
         G = grad_out[0]  # [N, S, O]
         I = module._inputs  # [N, S, I/q]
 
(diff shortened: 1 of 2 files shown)


bergson/gradients.py
```diff
@@ -442,15 +443,21 @@ def _save_input(self, module: nn.Module, inp: tuple, _):
         # to save memory, rather than waiting until the backward pass.
         p = self.processor.projection_dim
         if p is not None and not isinstance(norm, AdamNormalizer):
-            i = module.in_features
+            # Handle both Linear and Conv1D modules
```
luciaquirke: Unnecessary comment

bergson/gradients.py
```diff
@@ -442,15 +469,15 @@ def _save_input(self, module: nn.Module, inp: tuple, _):
...
+            i = getattr(module, LayerAdapter.in_func(module))
             x = x @ self.projection(name, p, i, "right", x.device, x.dtype).T  # type: ignore
 
         module._inputs = x
 
     def _process_grad(self, module: nn.Module, _, grad_out):
         """Process the incoming gradient wrt the output of the module."""
         # Sanity checks
-        assert isinstance(module, nn.Linear), "Expected a Linear module"
+        assert isinstance(module, LayerAdapter.supported_modules), "Expected a supported module"
```
luciaquirke: Could we print the supported modules here?

luciaquirke (approved):
LGTM 🚀

PR #54, opened 2025-10-23
[wip] Add support for including bias gradients

zywilliamli:
tested manually with both models with bias, and without bias, with the config on and off. 

this pr adds a configuration to include bias gradient, when enabled, on each module that includes bias, we will append the bias gradient as an extra column before projecting. (similar to the AdamNormalizer, this involves doing both sides of the projection on the backwards pass)


bergson/gradients.py
```diff
@@ -465,10 +484,14 @@ def _save_input(self, module: nn.Module, inp: tuple, _):
 
             x = x * b.type_as(x)  # [N, S, I] * [I] → [N, S, I]
 
+        include_bias = (
+            getattr(module, "bias", None) is not None and self.processor.include_bias
+        )
+
         # If we're not using AdamNormalizer, we can randomly project the input here
         # to save memory, rather than waiting until the backward pass.
```
luciaquirke: Update this comment to explain the reasoning behind the bias handling here (i.e. different gradient calculations being concatenated)

PR #55, opened 2025-10-23
clear hook lists so that they don't grow after each step

### LINKED ISSUE #42: ETA creep for large index builds
**Description**

ETA creep in the static index build that disappears when using the streamed and sharded index build settings with the same dataset. Structure in the batch ordering has been ruled out by shuffling the batches

**Reproduce**

```
uv run -m bergson runs/10b-test --model HuggingFaceTB/SmolLM2-135M --dataset EleutherAI/SmolLM2-135M-10B --skip_preconditioners  --truncation
```

In this run the ETA starts around 24 hours then increases to 82 hours after 2:43 hours of build time. A streaming version of this runs takes around 24 hours.

**TODO**

- [ ] Run a sharded index build without streaming mode to rule out an issue with large non-streaming datasets 
- [ ] Run a sharded index build where a subset of a single mmap is pass in for each shard to rule out issues with large mmaps

LouisYRYJ:
This PR fixes #42 
Specifically the isue was that GradientCollector was accumulating the hooks in its self._fwd_hooks and self._bwd_hooks and it took increasingly longer as we were calling .remove() on more and more entries (apparently .remove() is torch internal functionality of torch.utils.RemovableHandle and does not do affect the list itself).

TODO:
I get different results when running with and without the modification, but seems like a floating point error issue

### DIFF
--- a/bergson/gradients.py
+++ b/bergson/gradients.py
@@ -155,9 +155,9 @@ class AdamNormalizer(Normalizer):
         and the factored second moments.
         """
         # We assume avg_sq is a square matrix of shape [O, I]
-        assert (
-            self.avg_sq.ndim == 2
-        ), f"Expected 2D tensor for avg_sq, got {self.avg_sq.ndim}D"
+        assert self.avg_sq.ndim == 2, (
+            f"Expected 2D tensor for avg_sq, got {self.avg_sq.ndim}D"
+        )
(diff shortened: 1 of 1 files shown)


luciaquirke:
Tested and working, fabulous thank you!! 🚀

PR #63, opened 2025-11-09
Distributed refactor

LouisYRYJ:
This PR rewrites distributed setup to be more flexible.

### DIFF
--- a/tests/test_attributor.py
+++ b/tests/test_attributor.py
@@ -13,13 +13,12 @@ def test_attributor(tmp_path: Path, model, dataset):
     cfg = IndexConfig(run_path=str(tmp_path))
     cfg.skip_preconditioners = True
 
-    kwargs = {
-        "model": model,
-        "data": dataset,
-        "processor": GradientProcessor(),
-        "cfg": cfg,
-    }
-    collect_gradients(**kwargs)
+    collect_gradients(
+        model=model,
+        data=dataset,
+        processor=GradientProcessor(),
+        cfg=cfg,
+    )
 
     attr = Attributor(tmp_path, device="cpu", unit_norm=True)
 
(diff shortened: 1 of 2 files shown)


tests/test_build.py
```diff
@@ -32,21 +32,35 @@ def test_build_e2e(tmp_path: Path):
             "--truncation",
         ],
         cwd=tmp_path,
+        capture_output=True,  # Add this
+        text=True,  # Add this to get strings instead of bytes
     )
 
-    assert result.returncode == 0
+    # Print the output to see what's failing
+    if result.returncode != 0:
```
luciaquirke: I fixed this in the other query PR, I think it's a bit different than this fix. Basically returncode is 0 as long as the subprocess itself is happy so you parse the output text unconditionally

bergson/build.py
```diff
@@ -2,56 +2,103 @@
...
+    """Handle data loading and preprocessing"""
+
+    data_str = cfg.data.dataset
+    if data_str.endswith(".csv"):
+        ds = assert_type(Dataset, Dataset.from_csv(data_str))
+    elif data_str.endswith(".json") or data_str.endswith(".jsonl"):
+        ds = assert_type(Dataset, Dataset.from_json(data_str))
+    else:
+        try:
+            ds = load_dataset(data_str, split=cfg.data.split, streaming=cfg.streaming)
```
luciaquirke: Need to add back the lost subset arg

tests/test_attributor.py
```diff
@@ -5,16 +5,21 @@
...
-    collect_gradients(
-        model=model,
-        data=dataset,
-        processor=GradientProcessor(projection_dim=16),
-        path=tmp_path,
-    )
+    cfg = IndexConfig(run_path=str(tmp_path))
+    cfg.skip_preconditioners = True
+
+    kwargs = {
```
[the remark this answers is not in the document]
luciaquirke: if you can't see the keys for the input args to functions in your IDE something is broken, they should show up in grey

luciaquirke:
I'm going to yolo merge this, happy to raise another PR if there are issues or we want to add the kwargs pattern back to the tests or something like that 

Thanks for working on this!!

PR #80, opened 2025-11-17
Add back optimizer-aware gradients

luciaquirke:
It's not clear why these were removed, adding back to complete the TrackStar replication feature set.


LouisYRYJ:
Does this have the bias fixes that Baber did in https://github.com/EleutherAI/bergson/pull/76 ? Or what is the relation between this branch and his? 
I guess we first merge this and do Baber's PR?

Rest LGTM

Issue #91, opened 2025-12-08
Issues with building gradients on gemma-3


mmulet:
I've been having some trouble with bergson when using google-gemma-3, (using the latest berson version, version 0.40.0, commit 453e36ef5258b5695eddf9621a4aff3566e5dc69)

super simple script to reproduce:
```python
from bergson import IndexConfig
from bergson.build import build
from bergson.data import DataConfig
from pathlib import Path
lora_path f"./input/26_lora_otter_match_paper/filtered_models/1520/checkpoint-1520"

otter_index_path = Path(f'./output/otter_index')
otter_index_path.mkdir(parents=True, exist_ok=True)

build(
    IndexConfig(
        run_path=str(otter_index_path),
        model=lora_path,
        data=DataConfig(
            dataset=f"{run_folder}/input/teacher_numbers.jsonl",
            prompt_column="prompt",
            completion_column="completion"
        ),
        token_batch_size=4096
        # token_batch_size=8192
    )
)
```
error messages:
```bash
error messages 
W1208 22:54:22.944000 206150 torch/multiprocessing/spawn.py:174] Terminating process 206731 via signal SIGTERM
E1208 22:54:23.470000 206150 torch/distributed/elastic/multiprocessing/api.py:742] failed (exitcode: 1) local_rank: 0 (pid: 206730) of fn: dist_worker (start_method: spawn)
E1208 22:54:23.470000 206150 torch/distributed/elastic/multiprocessing/api.py:742] Traceback (most recent call last):
E1208 22:54:23.470000 206150 torch/distributed/elastic/multiprocessing/api.py:742]   File "/mnt/ssd-1/soar-data_attribution/mike/new_home/influence-animal-numbers/mike/emergent-misalignment/.venv/lib/python3.11/site-packages/torch/distributed/elastic/multiprocessing/api.py", line 697, in _poll
E1208 22:54:23.470000 206150 torch/distributed/elastic/multiprocessing/api.py:742]     self._pc.join(-1)
E1208 22:54:23.470000 206150 torch/distributed/elastic/multiprocessing/api.py:742]   File "/mnt/ssd-1/soar-data_attribution/mike/new_home/influence-animal-numbers/mike/emergent-misalignment/.venv/lib/python3.11/site-packages/torch/multiprocessing/spawn.py", line 220, in join
[...]

PR #93, opened 2025-12-13
add filter_modules config option

mmulet:
see issue #91


mmulet:
Moved it all the the collector and tested it. Works great!

PR #100, opened 2025-12-19
VRAM-efficient multi-GPU and/or multi-node preconditioner computation

luciaquirke:
More VRAM efficient variant where preconditioners can be spread across an arbitrary number of nodes to compute large outer products. This is useful because preconditioners are often applied to a query and then the query is run across a large dataset, so slow but VRAM-efficient preconditioner computation and usage is a scalable pattern.

The gradients computed from each data point on one device needs to be sent to all the other devices for the preconditioners to be updated, so this is not a drop-in replacement for our regular gradient collector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ We view attribution as a counterfactual question: **_If we "unlearned" this trai
 
 # Announcements
 
+**January 2026**
+- [Experimental] Support distributing preconditioners across nodes and devices for VRAM-efficient computation through the GradientCollectorWithDistributedPreconditioners. If you would like this functionality exposed via the CLI please get in touch! https://github.com/EleutherAI/bergson/pull/100
+
 **October 2025**
 - Support bias parameter gradients in linear modules: https://github.com/EleutherAI/bergson/pull/54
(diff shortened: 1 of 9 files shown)


bergson/collector/multi_node_gradient_collector.py
```diff
@@ -0,0 +1,412 @@
...
+    AdamNormalizer,
+    LayerAdapter,
+)
+from bergson.process_preconditioners import process_preconditioners
+from bergson.score.scorer import Scorer
+from bergson.utils.utils import assert_type
+
+
+@dataclass(kw_only=True)
+class MultiNodeGradientCollector(HookCollectorBase):
```
LouisYRYJ: is this going to be a replacement for GradientCollector? It seems like we don't it, if we have this one

luciaquirke: Also this does a distributed operation with the data every step so all the preconditioners get all the data so it will probably be too slow to be our main collector. It's mostly aimed at collecting big preconditioners where you only need to process a small amount of data to get a reasonable estimate. I guess it will be equally fast if you skip the preconditioners but slower in a scenario where you could fit all the precs on the same rank

Issue #107, opened 2025-12-27
load_gradient_dataset is very slow


LouisYRYJ:
https://github.com/EleutherAI/bergson/blob/8f9f6faceee42252172945d884374fc63bef8a17/bergson/data.py#L279

load_gradient_dataset should be a kind of 0 overhead functionality that simply binds gradients.bin entries with their corresponding hf dataset entries. This is not the case (e.g. we call .copy() on the mmap).
As far as I can see, we also don't really need this function? Maybe I am missing what it is needed for compared to load_gradients.
@luciaquirke

luciaquirke:
There was some bug without the copy that I unfortunately don't recall

PR #108, opened 2025-12-27
Reduce score nan fix

LouisYRYJ:
This PR has three bug fixes (probably should have done them seperately, but some were minor and I lost track):

- When rewards had a NaN, they would not be detected. Added logic to capture that and throw an error with an additional --skip_nan_rewards flag to simple remove entries with NaN rewards (see worker_utils.py)
- Use load_gradients instead of load_gradient_dataset ins score.py. load_gradient_dataset is very slow and also not necessary (see also https://github.com/EleutherAI/bergson/issues/107)
- When calling bergson reduce, we were doing cpu offloading which is extremely slow/inefficient and unnecessary. Keeping the gradients on the GPU memory (which usually should be no problem since it is the size of only 1 gradient) improves performance by a factor of ~70 from ~56s/it to ~0.79s/it (on some test model/dataset that I tried, will be worse on bigger datasets I think). 

Note: This is still WIP and I need to do final tests and clean up, but putting this out here to discuss and also flag that I am on this

### DIFF
--- a/bergson/score/score.py
+++ b/bergson/score/score.py
@@ -195,7 +195,18 @@ def get_query_ds(score_cfg: ScoreConfig):
     if not score_cfg.modules:
         score_cfg.modules = target_modules
 
-    mmap = load_gradients(Path(score_cfg.query_path), structured=False)
+    try:
+        mmap = load_gradients(Path(score_cfg.query_path), structured=False)
+    except ValueError as e:
+        if "integer won't fit into a C int" not in str(e):
+            raise e
+
+        print(
+            "Query gradients are too large to load with structure. "
+            "Attempting to load without structure..."
+        )
+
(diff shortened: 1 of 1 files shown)


luciaquirke:
Will take a look soon, I also have some changes coming up in https://github.com/EleutherAI/bergson/compare/trackstar-run?expand=1

mostly around multi-node runs and a data structure update for the preconditioners to support cases where you can't fit them all in CPU RAM simultaneously (converging on using memmaps for ~everything)

luciaquirke:
Since this is high priority work I just did a final review and actioned on it in this commit (just removing extraneous code), and I think this PR is now ready to merge: https://github.com/EleutherAI/bergson/pull/108/changes/6a548578a593fad8a93bd7baffeb66cbca3edf81

Please let me know if I can go ahead and merge it!

Issue #112, opened 2026-01-07
Numerical instability: outlier gradients leading to non-reproducible results


davidoj:
In the `data-args` branch, I've created the script `examples/preconditioner_instability.py`. This script generates two toy datasets, then computes gradients for each independently and both combined with bergsen build. There's significant variation in  gradient standard deviations for the same sets of datapoints between the indexes build in this way. This seems to be driven by rare numerical instability driven outliers. I'm not totally sure it's numerical instability, but in any case the outlier gradients aren't stable - they can occur only when we compute the index on the split dataset, or only when we compute the index on the combined dataset.

Filtering gradients >10x median seems to improve things signficantly.

Here's the script:
```python
"""
Reproduce gradient scale differences when computing gradients separately vs together.

Run with: python examples/preconditioner_instability.py
"""

import subprocess
from pathlib import Path

import torch
from datasets import Dataset

from bergson.data import load_gradients


def run_bergson_build(index_path: str, dataset_path: str):
    # Skip if index already exists
    if Path(index_path).exists() and (Path(index_path) / "info.json").exists():
        print(f"Index {index_path} already exists, skipping...")
        return

    cmd = [
        "bergson", "build", index_path,
        "--model", "Qwen/Qwen3-4B",
        "--dataset", dataset_path,
        "--prompt_column", "text",
        "--fsdp",
        "--projection_dim", "16",
        "--token_batch_size", "4000",
    ]
    print(f"Running: {' '.join(cmd)}")
    subprocess.run(cmd, check=True)


def main():
    # Create a simple dataset - just split some text in two
    # Need enough samples for 8 workers to each get at least one batch
    texts_a = [f"The quick brown fox jumps over the lazy dog {i}" for i in range(2000)]
    texts_b = [f"A journey of a thousand miles begins with a single step {i}" for i in range(2000)]

    ds_a = Dataset.from_dict({"text": texts_a})
    ds_b = Dataset.from_dict({"text": texts_b})
    ds_combined = Dataset.from_dict({"text": texts_a + texts_b})

    # Save datasets
    ds_a.save_to_disk("runs/instability_test/data_a")
    ds_b.save_to_disk("runs/instability_test/data_b")
    ds_combined.save_to_disk("runs/instability_test/data_combined")

    # Build three indices
    print("=" * 70)
    print("Building index A (first half)...")
    run_bergson_build("runs/instability_test/a", "runs/instability_test/data_a")

    print("=" * 70)
[...]

PR #117, opened 2026-01-09
Use dtype utils

luciaquirke:
Our dtype strategy has been haphazard. New strategy:

- Convert all losses to float32. This is the ubiquitous loss dtype.
- In `Scorer` convert gradients and queries to float32 by default. Provide other options (same as gradient dtype or arbitrary conversion) using a config parameter.
- In `build`, use the model gradient dtype (same as parameter dtype). Where multiple dtypes are used, take the highest precision one and log a warning.
- Add utils for dtype handling.
- Handle saving bfloat16s to np.memmap (there's a small library for this from the Jax maintainers, ml dtypes)

### DIFF
--- a/bergson/collector/collector.py
+++ b/bergson/collector/collector.py
@@ -565,7 +565,7 @@ def fwd_bwd_factory(cfg: IndexConfig) -> Callable:
             if "advantage" in batch:
                 losses *= torch.tensor(batch["advantage"], device=losses.device)
 
-        losses.mean().backward()
+        losses.sum().backward()
         model.zero_grad()
 
         return losses


luciaquirke:
@claude open a standalone PR with the fix changing loss from mean to sum, and add a test based on David's issue replication script to show that the gradients from his two builds are now identical. David's replication script is in the data-args branch, documented in https://github.com/EleutherAI/bergson/issues/112

luciaquirke:
@claude I gave you permissions, can you get this done now?

luciaquirke:
@claude try again?

luciaquirke:
@claude thanks, now could you remove that fix from this PR so it's a standalone dtype update PR?

PR #122, opened 2026-01-13
Add benchmarking code

luciaquirke:
The benchmarking code is in a subpar state but I am going to merge it anyway because:
- The code is generally self-contained so shouldn't affect the core library
- People are asking about expected runtimes and we can provide it

Related library changes (merged in previous PR):
- Add in-memory gradient collector for small runs
- Add auto batch size tool to CLI
- Include build metadata in reduce runs (we may want to update the library so this isn't necessary in the future)
- Support max_tokens with Dataset


luciaquirke:
@claude great review, can you fix the problems you found?

LouisYRYJ:
Sounds good, benchmarking seems pretty isolated so seems fine.
Bit confused about the projection_dim config? Can I still pass projection_dim=0 if I want to? And will that still return the expected behavior (no projection)?

PR #124, opened 2026-01-14
[WIP] Remove transformers pin, use cosine similarity in test instead of elementwise comparison

simonmnash:
Removes the transformers<4.56.0 pin by using cosine similarity and relative magnitude for normalized gradient comparison. test_gradients.py now passes with transformers 4.57.

WIP Element: Still doing some reading/experimenting to make sure the selected thresholds are appropriate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     "pyarrow",
     "simple-parsing",
     "torch",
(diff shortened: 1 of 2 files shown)


luciaquirke:
Hello, thanks for this! Any idea of what changed in the new transformers version that made the original error larger?

PR #130, opened 2026-01-22
Fix maximum value check for token_batch_size to allow larger values

smarter:
Previously, token_batch_size was limited by max_position_embeddings, but this is the maximum length of one sequence, and a batch can contain multiple sequences. As far as I can tell, the real limit is the default value of model_max_length in the tokenizer (some models have model_max_length == max_position_embeddings which causes this confusion).

We also make the truncation logic more explicit by passing a max_length parameter instead of mutating model_max_len.


luciaquirke:
LGTM + thanks very much for the fix! let's merge when conflicts are resolved

luciaquirke:
@claude could you please fix the merge conflicts

PR #132, opened 2026-01-28
Preconditioner/loss comparison for removing unwanted signals

davidoj:
Add an experiment that compares various methods for attributing "semantic" information but not "style" information.


luciaquirke:
This is fabulous, thank you so much! I updated the docs file to have more of a quickstart vibe. I'm pretty new to `skills` files so idk if I can review them effectively but it would be nice to put the ones specific to this experiment into a subdirectory because I can imagine the directory getting pretty hard to navigate quickly. Otherwise LGTM

luciaquirke (approved):

PR #134, opened 2026-02-02
[WIP] Enable per-token grad collection

luciaquirke:
Future work:

- data.py has too much content and needs decomposition
- See if we can have less code replication in the gradient collectors, document findings for follow up work
- Add usage example


LouisYRYJ:
sweet, will take a look today

PR #138, opened 2026-02-09
Fixes the OOM when using normalizers and preconditioners

### LINKED ISSUE #137: Running with normalizers and preconditioners OOMs
While trying to estimate the normalizer, the collector will also try to compute the preconditioner. Because normalizers don't use projection, the preconditioner will be huge and it blows up.

SrGonao:
Fixes #137


bergson/normalizer/fit_normalizers.py
LouisYRYJ: I think we can cut all of the normalizer stuff here, because we don't need a normalizer to compute the normalizer

PR #143, opened 2026-02-16
Add per-token gradient collection; bring in-memory collector to feature parity

luciaquirke:
This is one PR rather than two because we use the in-memory gradient collector as a sane way to test + write examples for the per-token collection and I really don't want to split it up


bergson/collector/gradient_collectors.py
```diff
@@ -158,7 +175,16 @@ def backward_hook(self, module: nn.Module, g: Float[Tensor, "N S O"]):
...
+                P = g.unsqueeze(-1) * a.unsqueeze(-2)
+                P = P.flatten(2).clamp_(self.lo, self.hi)  # [N, S, grad_dim]
+
+                # Filter to valid positions only
+                # Mask is [N, S]
+                P = P[self._current_valid_mask]  # [total_valid, grad_dim]
+            else:
+                P = g.mT @ a  # [N, O/p, S] @ [N, S, I/q] → [N, O/p, I/q]
 
         P = P.flatten(1).clamp_(self.lo, self.hi)
```
luciaquirke: TODO check

bergson/data.py
```diff
@@ -259,9 +749,13 @@ def load_data_string(
...
     if data_str.endswith(".csv"):
-        ds = assert_type(Dataset, Dataset.from_csv(data_str))
+        ds = Dataset.from_csv(data_str)
     elif data_str.endswith(".json") or data_str.endswith(".jsonl"):
-        ds = assert_type(Dataset, Dataset.from_json(data_str))
+        ds = Dataset.from_json(data_str)
+    elif Path(data_str).is_dir() and (Path(data_str) / "dataset_info.json").exists():
+        ds = Dataset.load_from_disk(data_str, keep_in_memory=False)
+        if isinstance(ds, DatasetDict):
+            ds = ds[split]
```
luciaquirke: I would like to exclude this code chunk from the review and take personal responsibility for it 🙏

PR #144, opened 2026-02-17
ci: run tests and typechecking in parallel

luciaquirke:
I think these got merged somehow, makes the build slow for no reason


luciaquirke:
Claude failure is expected because I'm changing its workflow file

PR #148, opened 2026-02-17
Add `preconditioners` and `trackstar` CLI commands

luciaquirke:
- Add `preconditioners` CLI command - wraps build() with skip_index=True to compute only normalizers and preconditioners. 
- Add `trackstar` CLI command - runs the full pipeline (preconditioner → preconditioner → reduce → score) as a single command, with separate --data.* and --query.* flags for value and query datasets.
- Update the Scores class to be a bit more user friendly

Includes example script scoring pile-10k against mean WMDP bio gradients.


bergson/utils/worker_utils.py
```diff
@@ -140,7 +140,6 @@ def setup_model_and_peft(
     try:
         peft_config = PeftConfig.from_pretrained(cfg.model)
     except ValueError:
-        print(f"PEFT config not found for model {cfg.model}")
```
luciaquirke: I think this is a stray debug print, it was being noisy in the example script output

PR #152, opened 2026-02-17
Remove reduce config; update docs

### LINKED ISSUE #142: Trackstar does not work correctly with preconditioners
We have to apply preconditioners before unit normalizing, so it has to happen in bergson reduce already. 
Also will need pipeline changes

luciaquirke:
Adds `docs/pipeline.rst` with explanations of what each CLI command does, what it produces, when to use each one, and how they fit together. Also includes a note on preconditioner ordering and a worked LESS-style example.

Closes #142


docs/pipeline.rst
```diff
@@ -0,0 +1,260 @@
...
+Choosing the Right Command
+--------------------------
+
+The decision tree below covers the most common scenarios:
+
+.. code-block:: text
+
+   Do you want to search a gradient index interactively (e.g. per-prompt)?
+   ├── Yes → use build + query
+   └── No  → Do you want to rank a dataset by influence against a fixed query?
```
luciaquirke: This question should be `Do you want to use full gradients without random projection or preconditioning`?

docs/pipeline.rst
```diff
@@ -0,0 +1,260 @@
...
+- You intend to query the index multiple times against different queries, so it's worth
+  paying the up-front cost to store all gradients.
+- You are using random projections (``--projection_dim > 0``) so each gradient is
+  small enough to store individually.
+
+**What it produces**
+
+A directory at ``run_path`` containing:
+
+- ``data.npy`` — a memory-mapped array of shape ``[num_examples, projection_dim]``
```
luciaquirke: the shape is wrong, it's a structured array

docs/pipeline.rst
```diff
@@ -0,0 +1,260 @@
...
+  (or ``[num_examples, param_dim]`` if no projection).
+- ``indices.jsonl`` — per-example metadata.
```
luciaquirke: wrong i think

PR #153, opened 2026-02-17
[Large Refactor] Enable gradient preprocessing in reduce and in score

### LINKED ISSUE #154: Enable grad preprocessing in reduce
similar to the preprocessing in score but per-gradient

luciaquirke:
Closes https://github.com/EleutherAI/bergson/issues/154
Closes https://github.com/EleutherAI/bergson/pull/158

- Support split preconditioning when unit normalize is true
- This will provide more flexibility around where preprocessing is applied and improve the documentation of gradient preprocessing

image


bergson/config.py
```diff
@@ -300,6 +312,18 @@ class ScoreConfig:
     unit_normalize: bool = False
     """Whether to unit normalize the gradients before computing the scores."""
 
+    accumulation: Literal["mean", "sum", "none"] | None = None
```
luciaquirke: @claude we don't need a string "none" if we already have a real None

bergson/collector/dist_preconditioners_gradient_collector.py
```diff
@@ -280,7 +280,7 @@ def teardown(self):
         # Flush and reduce builder if it exists
         if self.builder is not None:
             self.builder.flush()
-            self.builder.dist_reduce()
+            self.builder.teardown()
```
luciaquirke: TODO put the flush in teardown

luciaquirke:
@claude can you add a docs page discussing query and index preprocessing and when each of them should happen? Give case studies of how you would handle a full-gradients cosine similarity example with an optimizer normalizer, a full-gradients inner product example with an optimizer normalizer, a randomly projected gradients example with reduce and score, and a randomly projected gradients example with unit normalize, preconditioners, build, and score.

luciaquirke:
@claude in the docs can you move the experiment walkthroughs and gradient preprocessing sections above the automatically generated docstrings sections

luciaquirke:
@claude 

> Random projections (Johnson-Lindenstrauss) approximately preserve inner products and cosine similarities while reducing gradient dimensionality by orders of magnitude.
Should be
> Random projections approximately preserve inner products and cosine similarities  (Johnson-Lindenstrauss) while reducing gradient dimensionality by orders of magnitude.

> reduce aggregates all query gradients into a single vector (mean or sum) without storing any per-example gradients.
Should be 
reduce aggregates all query gradients into a single vector (mean or sum) without storing per-example gradients.

> Optimizer normalization must be applied during gradient collection (set --normalizer at both reduce and score time). It cannot be applied after the mean-reduction in reduce, since applying the normalizer to the mean gradient is not the same as normalizing each gradient then taking the mean.
Should be 
> Optimizer normalization must be applied during gradient collection (set --normalizer at both reduce and score time). It cannot be applied after the mean-reduction in reduce - the normalizer is non-linear so applying it to the mean gradient is not the same as normalizing each gradient then taking the mean.

> When query and index datasets come from different distributions, --mixing_coefficient (default 0.99) interpolates between their second moment matrices:
Should be
> When query and index datasets come from different distributions, --mixing_coefficient (default 0.99) interpolates between their second moment matrices (i.e. the empirical Fisher information matrices):

> Adjust this when the query dataset is small (causing noisy 𝐻query estimates) or when the query and index distributions diverge significantly.
Should be
> Adjust this according to the guidelines in https://arxiv.org/abs/2410.17413

Otherwise good

bergson/config.py
```diff
@@ -265,6 +265,23 @@ class QueryConfig:
     its top results as rows with columns: query, result, result_index, score."""
 
 
+@dataclass
+class PreprocessConfig:
```
LouisYRYJ: I am not sure about this current version of the config, it is very focused on trackstar.

luciaquirke: TODO prompt: could you please pull out the preconditioner mixing logic and configuration into a standalone function we use in the TrackStar CLI command? we probably want to save the mixed preconditioner to disk and then pass the path into the next step. We may need to save some config.json type files to the mixed preconditioner directory to facilitate loading

luciaquirke:
@claude could you please address the last three comments

bergson/score/scorer.py
```diff
@@ -52,10 +67,32 @@ def __init__(
...
+        if self.preconditioners:
+            offset = 0
+            for m in modules:
+                d = query_grads[m].shape[1]
+                if m in self.preconditioners:
+                    self.query_tensor[:, offset : offset + d] = (
+                        self.query_tensor[:, offset : offset + d].float()
+                        @ self.preconditioners[m].float()
+                    ).to(self.dtype)
+                offset += d
```
LouisYRYJ: what's the reason to do this rather than just skipping the cat and just looping through the dict keys?

bergson/data.py
```diff
@@ -321,16 +331,29 @@ def __init__(
...
         )
 
+    def _compute_h_inv(self, device: str | torch.device) -> dict[str, torch.Tensor]:
+        """Compute preconditioner from preprocess_cfg if available."""
+        if self.preprocess_cfg is None:
+            return {}
+        device = torch.device(device)
+        return compute_preconditioner(
+            self.preprocess_cfg.preconditioner_path,
+            power=-0.5 if self.preprocess_cfg.unit_normalize else -1,
```
LouisYRYJ: see above

bergson/score/scorer.py
```diff
@@ -1,12 +1,18 @@
...
 from bergson.score.score_writer import ScoreWriter
 
 
 class Scorer:
     """
     Scores training gradients against query gradients.
 
+    Handles all preconditioning internally:
+      - Loads preconditioner from disk if ``preconditioner_path`` is given.
+      - Applies to query grads once at init time.
```
LouisYRYJ: Hmmh how about we completely remove query preprocessing completely from the Scorer? Then it either happens in bergson reduce, build, or in a custom function?

luciaquirke: I feel a little sad about giving people 3+ commands to call to get one set of results. I guess we could define a bunch of CLI commands for different DA algorithms and make the current layer a lower-level thing? This seems like a fairly big chunk of work though

bergson/score/scorer.py
```diff
@@ -52,10 +67,32 @@ def __init__(
...
+        # Load preconditioner: H^(-1/2) for split, H^(-1) for one-sided
+        self.preconditioners = compute_preconditioner(
+            preconditioner_path,
+            device=device,
+            power=-0.5 if unit_normalize else -1,
+        )
+        if self.preconditioners:
+            self.preconditioners = {
+                k: v.to(dtype=dtype) for k, v in self.preconditioners.items()
+            }
```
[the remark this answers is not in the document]
luciaquirke: TODO add `dtype` to function signature

bergson/process_grads.py
```diff
@@ -0,0 +1,252 @@
...
+
+def normalize_grad(
+    grad_dict: dict[str, torch.Tensor],
+    unit_normalize: bool,
+    device: torch.device,
+) -> dict[str, torch.Tensor]:
+    """Preprocess a single gradient. Optionally unit-normalizes
+    across all columns, moves to device."""
+    grads = {
+        name: g.to(device=device, dtype=torch.float32) for name, g in grad_dict.items()
```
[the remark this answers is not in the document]
luciaquirke: Sometimes the original gradients and the normalized gradient can all be represented in another dtype but converting from one to the other adds a lot of error. I think we should cast it back to the original `dtype` at the end though.

bergson/process_grads.py
```diff
@@ -0,0 +1,252 @@
...
+    if power == -0.5:
+        return {
+            name: psd_rsqrt(H.to(device=device, dtype=torch.float32))
+            for name, H in preconditioners.items()
+        }
+    elif power == -1:
+        return {
+            name: compute_damped_inverse(H.to(device=device))
+            for name, H in preconditioners.items()
+        }
```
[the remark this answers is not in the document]
luciaquirke: Sorry I'm still finding the right balance with the coding agents

bergson/score/scorer.py
luciaquirke: I think I can replace these all with dict + no cat but I'll write a little benchmark

bergson/process_grads.py
```diff
@@ -0,0 +1,252 @@
...
+    i_proc = GradientProcessor.load(index_path)
+
+    mixed_preconditioners = {
+        k: q_proc.preconditioners[k] * mixing_coefficient
+        + i_proc.preconditioners[k] * (1 - mixing_coefficient)
+        for k in q_proc.preconditioners
+    }
+
+    # Build a new processor with the mixed preconditioners
+    mixed_proc = GradientProcessor(
```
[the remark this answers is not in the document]
luciaquirke: Yes I guess so, but maybe in a follow up PR? This one is already monstrous

bergson/query/attributor.py
```diff
@@ -87,6 +90,22 @@ def __init__(
 
         self.ordered_modules = mmap.dtype.names
 
+        # Precompute preconditioners
```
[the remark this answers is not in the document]
LouisYRYJ: hmmh also this is again very trackstar specific, it's fine for now to unblock us I think, but let's flag it so we know about it

bergson/score/score.py
```diff
@@ -98,16 +98,16 @@ def preprocess_grads(
         norms = torch.cat(list(grads.values()), dim=1).norm(dim=1, keepdim=True)
         grads = {k: v / norms for k, v in grads.items()}
 
-    # Accumulate across items
-    if accumulate_grads == "mean":
+    # Aggregate across items
+    if aggregate_grads == "mean":
```
LouisYRYJ: why do we keepdim=True here?

bergson/score/scorer.py
LouisYRYJ: Yes, I think we should make it as easy as possible for users to "just do TrackStar". Users who want to have more custom functions can then read our constructed scripts and modify as desired.

bergson/score/scorer.py
```diff
@@ -52,33 +79,62 @@ def __init__(
...
-        # Convert the gradients to the scoring dtype
-        if next(iter(mod_grads.values())).dtype != self.dtype:
-            mod_grads = {name: grad.to(self.dtype) for name, grad in mod_grads.items()}
-
         scores = self.score(mod_grads)
         self.writer(indices, scores)
 
     @torch.inference_mode()
-    def score(self, mod_grads: dict[str, torch.Tensor]) -> torch.Tensor:
+    def score(self, index_grads: dict[str, torch.Tensor]) -> torch.Tensor:
```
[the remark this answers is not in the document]
LouisYRYJ: I am fine to do this in a new PR, but we should flag it at all the places relevant

LouisYRYJ:
This looks pretty solid 🚀 
I think after we go through the current round of comments, we can merge it!

bergson/score/scorer.py
```diff
@@ -52,33 +79,62 @@ def __init__(
...
-            dim=1,
+        # Load preconditioner: H^(-1/2) for split, H^(-1) for one-sided
+        self.preconditioners = get_trackstar_preconditioner(
+            preconditioner_path,
+            device=device,
+            power=-0.5 if unit_normalize else -1,
+            return_dtype=dtype,
         )
+        # Precondition query grads per module, then cat into a single tensor
+        if self.preconditioners:
```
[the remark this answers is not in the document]
luciaquirke: I guess we could add another path that doesn't cost memory, I think this is fast though

bergson/score/score.py
```diff
@@ -98,16 +98,16 @@ def preprocess_grads(
...
         grads = {name: grads[name].sum(0, keepdim=True) for name in grad_column_names}
-    elif accumulate_grads != "none":
-        raise ValueError(f"Invalid accumulate_grads: {accumulate_grads}")
+    elif aggregate_grads != "none":
+        raise ValueError(f"Invalid aggregate_grads: {aggregate_grads}")
 
-    # Normalize the accumulated gradient
-    if normalize_accumulated_grad:
+    # Normalize the aggregated gradient
+    if normalize_aggregated_grad:
```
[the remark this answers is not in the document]
luciaquirke: This is not a hot loop and should happen before model loading so shouldn't be memory intensive either. I think this way is faster but more memory intensive (bench results below)

PR #158, opened 2026-02-23
Fixes unit normalizing when applying a preconditioner, following trackstar

SrGonao:


bergson/score/scorer.py
```diff
@@ -73,6 +76,13 @@ def __call__(
     @torch.inference_mode()
     def score(self, mod_grads: dict[str, torch.Tensor]) -> torch.Tensor:
         """Compute scores for a batch of gradients."""
+        if self.preconditioners:
+            mod_grads = {
+                name: grad @ self.preconditioners[name].to(self.device)
+                for name, grad in mod_grads.items()
+            }
+        # TODO: do we need to have both grads and mod_grads in memory at the same time?
```
luciaquirke: You could do torch.cat on mod_grads first and then apply the preconditioners to the relevant slices of the concatenated tensor using an offset + for loop, if you search "offset" in the repo there should be an analogous pattern in there

PR #162, opened 2026-02-25
Fix some test set leaks in preconditioner comparison experiment

davidoj:
- Some of the preconditioner comparison measures were leaking test set data to the preconditioner
- Added a "semantic PCA" style ablation, like the ordinary style ablation but using gradients masked to "semantic" tokens (it's worse than the regular style ablation)
- Added a report on variance explained by PCA


luciaquirke (approved):
(Scanned and looks reasonable)

PR #163, opened 2026-02-26
Fix attention head splitting bug and add module name validation

luciaquirke:
Update `module._inputs` handling so it doesn't break when the `split_attention_heads` wrapper runs, fixing a bug where the cached activations were deleted after the first head. Add test.

Add early validation in `HookCollectorBase.__post_init__` that raises an error when `attention_cfgs` keys don't match any model module names.

Fix API drift in the induction example (HeadConfig→AttentionConfig, head_cfgs→attention_cfgs, path typing, query reduction args) and correct test module names so head splitting is actually exercised.


luciaquirke:
@LouisYRYJ the important change here is extracting the del module._inputs into shared logic

PR #167, opened 2026-03-02
fix: convert PyArrow Column to list before allocate_batches everywhere

luciaquirke:
HuggingFace Dataset column access (ds["length"]) returns a PyArrow Column, not a Python list. Iterating over it element-by-element (via sorted(), random indexing) is ~1000x slower than on a native list.


luciaquirke:
TODO update claude max turns to 40 (~$1.60)

PR #174, opened 2026-03-06
Add MAGIC CLI with runtime DTensor double-backward patch

luciaquirke:
## My Changes Summary

- Add MAGIC CLI. Brings in DataConfig to start giving us the features the other Bergson CLIs have. Will likely need more work over time to bring to feature parity. 
- The query reduction currently only works if the number of queries divides nicely across the world size. I have a fix for this in a follow-up PR but this current one was getting ungainly.

## Things which could go in this PR or a follow-up

- Save config jsons to disk 
- Use gradcheck to test mixed precision? Or any other ideas for correctness tests?
- no_dist is only available on later torch versions (2.9ish+) 
- support reading configuration from yaml so it's easier to document configurations
- per-token scores (PR exists)
- padding for batches that don't fit nicely into world size (PR exists) 

## Claude Changes Summary

- **Runtime DTensor patch** (`bergson/magic_patch.py`): Monkey-patches `Redistribute.backward` and `_ToTorchTensor.backward` at runtime to make FSDP redistribution twice-differentiable (pytorch/pytorch#160509). Replaces the old `magic_wmdp_setup.sh` that modified torch source files on disk. Idempotent — call `apply_dtensor_patch()` before any DTensor double-backward operations.

## Test plan

- [x] End-to-end MAGIC attribution run on GPU cluster


bergson/double_backward.py
```diff
@@ -0,0 +1,325 @@
...
+    model: str = "EleutherAI/pythia-160m"
+    """HuggingFace model name."""
+
+    revision: str | None = None
+    """Model revision (branch, tag, or commit hash)."""
+
+    data: DataConfig = field(default_factory=DataConfig)
+    """Training dataset."""
+
+    query: DataConfig = field(default_factory=lambda: DataConfig())
```
norabelrose: nitpick: you should write `field(default_factory=DataConfig)` like you did above lol

luciaquirke:
I think once we can replicate the good spearman correlations with the CLI it's basically mergeable

PR #188, opened 2026-03-10
[WIP] Skip gradient collection for documents with fewer than two tokens

luciaquirke:
Documents with fewer than 2 tokens cannot produce valid next-token labels, and length-0 documents create [N, 0] tensors that hang the model forward pass. In multi-GPU settings, the bin-packing allocator assigned all zero-length documents to a single rank (cost = 0), causing that rank to stall while others completed their NCCL all-reduces.

Fix by filtering <2-token documents from batch allocation in _allocate_batches_world. Their gradient index entries remain at the pre-initialized zero value, preserving the dataset-to-score index mapping. 

TODO:

we probably want to update the  "is_written" column in scores at these positions to reflect the fact that the 0. gradient is intentional. We may also want to raise an error here unless a --skip_empty_docs flag is given.

### DIFF
--- a/bergson/data.py
+++ b/bergson/data.py
@@ -244,10 +244,27 @@ def _allocate_batches_world(
     """
     if ranks is None:
         ranks = list(range(world_size))
-    if len(doc_lengths) < world_size:
-        raise RuntimeError("Not enough documents to distribute across workers.")
-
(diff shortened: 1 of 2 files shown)


norabelrose (approved):
LGTM

PR #189, opened 2026-03-11
[WIP] Add mixed precision gradient collection

luciaquirke:
- Always leave optimizers and preconditioners in fp32
- Use mixed precision fwd-bwd

My conclusions from the test below are basically:
- In practice this makes retraining effects in either fp32 or bf16 essentially indistinguishable
- TrackStar in its current form is lame for coreset selection
- Eval selection results TODO
- Single item-based selection results TODO

**Coreset selection** (last regression: https://github.com/EleutherAI/bergson/pull/178):

I tried to repeat the analysis from the last PR but I became suspicious of the results because the training loss drop on this dataset sucks, because the model has already undergone IFT. So I did a second round using Qwen + LoRA + a dataset Qwen's bad at (sheet music) so get a smooth loss drop:

```
 torchrun --nproc_per_node 8 -m examples.filter_data \
    --model Qwen/Qwen2.5-1.5B \
    --dataset sander-wood/irishman \
    --prompt_column "abc notation" \
    --max_samples 10000 \
    --num_examples 1000 \
    --num_epochs 1 \
    --precision fp32 \
    --learning_rate 5e-5 \
    --subset default \
    --filter random 
```

I used an effective batch size of 128, LoRA rank 64, projection dim 16  and took the query over the entire dataset (should not have an eval set for coreset)

**ordered=False Leave-90%-out coreset selection (train loss), seed=42, 10k full dataset** (works, narrowly)

"Attribution": 2.77 (raw grad cosine sim)
Random: 2.52
TrackStar BF16: **2.38**
TrackStar FP32: 2.43
Full dataset: 1.77

**ordered=True Leave-90%-out coreset selection (train loss), seed=42, 50k full dataset** (works, narrowly)

Full dataset: 1.287
Random: 1.878
TrackStar BF16: **1.763**
TrackStar FP32: **1.763**

**ordered=True Leave-80%-out retrain train loss, seed=42, 50k full dataset** (actually kind of sucks because loss does a big spike at the end for trackstar fp16 and bf16, which are almost exactly the same. probably they work better if you can maintain stability still but the gains here aren't large enough for me to keep experimenting).

Full dataset: 1.287
Random: 1.52
TrackStar BF16: 1.58
TrackStar FP32: 1.57

I tried the eval set query filtering, one step closer to the core thing TrackStar is supposed to be good at:

**ordered=True Leave-90%-out retrain eval loss, seed=42, 50k full dataset**

Full dataset:
Random:
TrackStar BF16:
TrackStar FP32:

image

Note that mixed precision attribution matches our mixed precision training so it may have some predictive advantage over single precision attribution.

TODO
- check we maintain correct dtype outside trackstar, in normalizers
- make coreset selection clean for future regression testing

### DIFF
--- a/bergson/config.py
+++ b/bergson/config.py
@@ -373,7 +373,7 @@ class FaissConfig:
 
 
 @dataclass
-class TrackstarConfig:
+class TrackStarConfig:
     """Config for the trackstar pipeline query dataset."""
 
     query: DataConfig = field(default_factory=DataConfig)
@@ -386,9 +386,14 @@ class TrackstarConfig:
     index preconditioners intersect at this component. Typical value is
     ~1000 out of ~65K total components."""
 
-    num_stats_sample_preconditioner: bool = True
-    """Whether to use num_stats_sample items or the full dataset to
-    compute preconditioners."""
+    sample_preconditioners: bool = True
+    """Whether to use num_stats_sample items to compute preconditioners.
+    If False, uses the full dataset."""
+
+    stats_token_batch_size: int | None = None
+    """Token batch size for normalizer/preconditioner estimation steps (1-2).
+    These components are always collected in fp32, so use ~2x the VRAM when
+    in half precision. If None, uses the main token_batch_size."""
 
     resume: bool = False
     """Skip pipeline steps whose output directory already exists."""
--- a/bergson/config.py
+++ b/bergson/config.py
@@ -157,6 +157,12 @@ class IndexConfig:
     processor_path: str = ""
     """Path to a precomputed processor."""
 
+    adam_state_path: str = ""
+    """Path to a training checkpoint directory containing an optimizer.pt
+    or to an optimizer state file. Extracts Adam exp_avg_sq second
+    moments and loads them into AdamNormalizer. Mutually exclusive with
+    fitting normalizers via --normalizer."""
+
     normalizer: Literal["adafactor", "adam", "none"] = "none"
     """Type of normalizer to use for the gradients."""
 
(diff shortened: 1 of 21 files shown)


bergson/normalizer/fit_normalizers.py
```diff
@@ -65,7 +65,7 @@ def adafactor_update(self, name: str, g: torch.Tensor):
         normalizer.col.add_(col_acc)
 
     def adam_update(self, name: str, g: torch.Tensor):
-        sq = g.square_().float().sum(0)
+        sq = g.float().square_().sum(0)
```
luciaquirke: I am really confused about why this was square then float

bergson/config.py
```diff
@@ -390,5 +390,10 @@ class TrackstarConfig:
     """Whether to use num_stats_sample items or the full dataset to
     compute preconditioners."""
 
+    high_precision_token_batch_size: int | None = None
```
luciaquirke: This name sucks but we want something like this

bergson/gradients.py
```diff
@@ -303,9 +303,10 @@ def normalize_(
         eps: float = 1e-8,
     ) -> Tensor:
         """Normalize the gradients by the square root of the second moments."""
-        # Adam-style epsilon is added outside the square root
-        denom = self.avg_sq.sqrt()
-        return grad.div_(denom.add_(eps))
+        # Adam-style epsilon is added outside the square root (in fp32 for precision),
+        # then cast to match grad dtype for the in-place division.
+        denom = self.avg_sq.sqrt().add_(eps).to(grad.dtype)
+        return grad.div_(denom)
```
luciaquirke: I'm not sure whether it matters but we can precompute this denom

PR #202, opened 2026-03-24
Replace fit_normalizers with loading from optimizer state

luciaquirke:
Load Adam second moments (exp_avg_sq) directly from training checkpoints instead of estimating them. Removes NormalizerCollector and fit_normalizers in favor of load_from_optimizer which reads optimizer.pt files.


bergson/utils/load_from_optimizer.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    state_path = Path(optimizer_state_path)
+    if state_path.is_dir():
+        state_path = state_path / "optimizer.pt"
+
+    optimizer_state = torch.load(state_path, map_location="cpu", weights_only=False)
+
+    # The optimizer state is keyed by position in the trainable parameter list.
+    # For LoRA checkpoints, only include LoRA params.
+    lora_params = [(n, p) for n, p in model.named_parameters() if "lora" in n]
```
norabelrose: hmmm we should be able to do this in a more robust way right? elsewhere we support all different kinds of `PeftModel`s, not just LoRAs

norabelrose (approved):
lgtm except the lora thing

PR #207, opened 2026-03-24
GPT-2 Fine-Tune Replication

luciaquirke:
- Add a learning rate scheduler config to bring the available schedules in line with HF (with the addition of lr_start in polynomial, required for the GPT-2 fine-tune).
- Add Trainer resume option for training phase (future work: backward phase)
- Drop data that doesn't fit evenly into the world size or batch size
- Add GPT-2 fine-tune replication yaml
- Fix a bug in distributed.py DTensor usage that pops up when models have tensors that aren't cleanly divisible across ranks (only applies to GPT-2 so far)

- Add validate_batch_size. This function may be useless because it does the same thing as the first training step. But we've had issues in the past that haven't been fully diagnosed where a too-large batch size in FSDP can cause a NCCL hang rather than an immediate crash. This can happen mid-training if the batch size is variable. If this happens it might be nice to have it happen in a dedicated location. I'm not sure if this will be relevant to us yet, need to look into our planned batching strategies.
 
Future work:
- Add fake data to make batches even without dropping data, and ignore their gradients

### DIFF
--- a/bergson/magic/cli.py
+++ b/bergson/magic/cli.py
@@ -1,9 +1,10 @@
+import math
 import os
 import shutil
 from dataclasses import asdict, dataclass
 from datetime import timedelta
 from pathlib import Path
-from typing import Literal
+from typing import Callable, Literal
 
 import torch
 import torch.distributed as dist
@@ -13,7 +14,6 @@ from scipy.stats import describe, spearmanr
 from simple_parsing import ArgumentParser, field
 from torch.distributed.tensor import init_device_mesh
 from torchopt.pytree import tree_iter
-from torchopt.typing import Numeric
 from tqdm import tqdm
 
 from ..config import AttributionConfig, DataConfig, TrainingConfig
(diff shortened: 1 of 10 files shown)


bergson/magic/cli.py
```diff
@@ -96,7 +100,88 @@ def compute_query_gradients(
...
+                return _warmup(step)
+            return lr
+
+    elif scheduler_type == "linear":
+
+        def _schedule(step):
+            if step < warmup_steps:
+                return _warmup(step)
+            progress = (step - warmup_steps) / decay_steps
+            return lr - (lr - lr_end) * progress
```
davidoj: I don't know if it practically matters very much but if you ever call get_schedule with num_steps < the actual number of training steps you take this will behave oddly (why? Maybe you want to bodgy a three-phase scheduler or something). Could clamp `progress = max(1, (step - warmup_steps) / decay_steps)` to avoid this

davidoj (commented):
Looks fine, left a couple of comments, address them at your discretion

norabelrose (changes requested):

davidoj (approved):
Looks good to me

PR #209, opened 2026-03-25
Add numerical stability diagnostic and fix double BOS in chat tokenization

davidoj:
## Summary
- **`--force_math_sdp` flag**: Disables flash/memory-efficient SDPA backends that cause inconsistent per-example gradients with different padding lengths
- **`bergson test_numerical_stability` command**: Automatically tests escalating configurations (default → force_math_sdp → fp32 → both) and reports minimum flags needed
- **Special token diagnostic**: Detects BOS/EOS duplication by comparing bergson's two-step chat tokenization against `apply_chat_template(tokenize=True)` ground truth. If `add_special_tokens=False` fails, checks whether `True` works and explains what to change
- **Fix**: Chat template tokenization path now uses `add_special_tokens=False` to prevent double BOS for models like LLama and Gemma

## Test plan
- [x] Full `bergson test_numerical_stability` on Pythia-160M (gradient consistency + special tokens)
- [x] Special token check on LLama 3.2 (detects double BOS), Gemma 2 (same), Qwen 2.5/3/3.5 (no BOS), Phi 4 (unused BOS), Pythia (no chat template — skips)
- [x] pre-commit passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,27 @@ bergson build <output_path> --model <model_name> --force_math_sdp
 bergson build <output_path> --model <model_name> --force_math_sdp --precision fp32
 ```
 
+### Performance impact
+
+The overhead of `--force_math_sdp` and `--precision fp32` varies by model. Benchmarked on A100-80GB with 500 documents from pile-10k:
+
+| Model | Settings | Build time | vs bf16 baseline |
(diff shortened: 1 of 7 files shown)


davidoj:
I'll get this right at some point (this was based on local main which was behind)

PR #210, opened 2026-03-25
YAML goodies

norabelrose:
You can now run `bergson <command> <config path>` to run that command with the arguments provided at that path. YAML and JSON are supported.


luciaquirke (approved):
Absolute cinema

PR #211, opened 2026-03-25
Diagnostic followup: show sample size, fix config label

davidoj:
## Summary
- Report now shows trial count in the summary header so readers know the sample size
- Renamed "defaults (precision=bf16)" label to "precision=bf16" — since the actual default precision is fp32, calling bf16 "defaults" was misleading

## Test plan
- [x] Verified output format with `DiagnoseConfig(model='EleutherAI/pythia-160m', n_trials=3)`

### DIFF
--- a/bergson/__main__.py
+++ b/bergson/__main__.py
@@ -154,11 +154,15 @@ class Score(Serializable):
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


luciaquirke (approved):
Thank you so much 🙏

PR #221, opened 2026-03-30
Adding padding examples instead of removing them in MAGIC

SrGonao:
Instead of removing examples at the end of the dataset I pad it with the last example but set the weights to 0. Was a bit unsure with what to do with num_train_docs, but think this is correct. Did not pad the queries because they dont use data weights and I think it is fine to drop some.

I also added dataset shuffling by mistake (maybe should have been a different pr) because I'm using it


norabelrose (changes requested):
We should also do this for the queries as well, I'll try to fix this

PR #227, opened 2026-04-01
Fix distributed hangs

norabelrose:
Batches were not getting padded properly, so that different ranks were seeing batches with different numbers of tokens

### DIFF
--- a/bergson/data.py
+++ b/bergson/data.py
@@ -582,6 +582,11 @@ def pad_and_tensor(
 
     # find max length
     max_len = max(len(seq) for seq in sequences)
+    if dist.is_initialized():
+        max_len_tensor = torch.tensor(max_len, device=device)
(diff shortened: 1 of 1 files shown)


luciaquirke:
@norabelrose hmm I'm still getting a hang. Not sure about this

luciaquirke:
If there is a large variance in sequence length so there are some 2-token items and some long ones you can get one device that's like [1, 2048] and another that's like [1024, 2] and then when you pad everything to the max length you get a batch that's like [1024, 2048] and everything hangs (crashes on one rank). I'm making this flag-controlled so we can keep your MAGIC fix and turn it off for everything else. But this isn't a great long term solution (due to my limited understanding of the relevant factors).

For MAGIC we can at least move the crashes upfront by finding the largest length-width discrepancy upfront.

PR #228, opened 2026-04-02
Update loss reduction default

luciaquirke:
This will more closely match TrackStar.

This will require any runs in EM coupled to Bergson's defaults to explicitly set `loss_reduction`="mean" to maintain reproducibility


norabelrose (approved):
LGTM

PR #239, opened 2026-04-19
Fix running KFAC with bfloat16 weights

smarter:
This fixes two issues when running a model with --precision bf16:

1. Computing `a.mT @ a` and `g.mT @ g` in bf16 is too inaccurate and leads to broken eigenvectors.

2. `torch.einsum` in ShardedMul raised `RuntimeError: expected scalar type Float but found BFloat16` when the stored eigenvectors (fp32) were combined with the model's bf16 activations/gradients.

Also get rid of --hessian_dtype "auto" and default to "fp32" since using less
precision doesn't seem to work well in practice.


luciaquirke (approved):
@LouisYRYJ might want to take a look but this all looks great to me, thank you!

LouisYRYJ:
@smarter Do you want to do the test editing in this PR?

smarter:
This is now tested, though one has to explicitly use --precision bf16 since the default is still fp32: https://github.com/EleutherAI/bergson/blob/c5171543a20fadd4ce2ae8febbadee1533b6a1c1/tests/ekfac_tests/conftest.py#L46

I'm wondering if for the CI we should test both fp32 and bf16 or just default to bf16 (could be done in a separate PR).

(Also Pythia-14m seems broken with --world_size 8 as described in the [last commit](https://github.com/EleutherAI/bergson/pull/239/changes/81cccfe8b6dea701857056b1929c920df9d4d91a) but I don't think that's news)

LouisYRYJ:
Thanks very much @smarter 
Very happy to see this in main now!

PR #241, opened 2026-04-23
Bugfixes and tests for per-token MAGIC

davidoj:
Added unit tests for per-doc and per-token equivalence: https://github.com/EleutherAI/bergson/pull/241#issuecomment-4311125028

Save doc_ids.pt after running magic in per-token mode, so we can easily map scores back to original docs: https://github.com/EleutherAI/bergson/pull/241#issuecomment-4331993851

Magic now more often respects splits/slices in the dataset config: https://github.com/EleutherAI/bergson/pull/241#issuecomment-4332187629

Magic now validates per-token if "per_token: true" is set (previously tried to validate per-doc and failed due to mismatched shapes between val vector and score array). https://github.com/EleutherAI/bergson/pull/241#issuecomment-4340209692


luciaquirke (approved):
Thank you 🙏

davidoj:
Added a few unit tests: `test_magic_per_token_scores_zero_at_masked_labels`,  `test_magic_per_token_sums_to_per_doc, test_magic_per_token_sums_to_per_doc_packed`,  `test_magic_per_token_sums_to_per_doc_with_padding`

Of these tests, all already passed except for  `test_magic_per_token_sums_to_per_doc_with_padding`
                                                                                                                                                                                                                                
Here's an example of the issue that test exercises:

We start with 3 chunks, each chunk is one doc, batch_size=2 → remainder=1, pad_count=1.

Original dataset (3 rows, num_docs=3):

 row 0: doc_ids = [0,0,0,0,0]
 row 1: doc_ids = [1,1,1,1,1] 
 row 2: doc_ids = [2,2,2,2,2]

In order to fill out the batch, we add a pad row:
 dataset.select([0, 1, 2,  2])                                                                                                                                                                                      

Previously, this would leave us with a copy of the last real document's doc_ids in the padding row:
                                                                                                                                                                                                                                  
 row 0: doc_ids = [0,0,0,0,0]
 row 1: doc_ids = [1,1,1,1,1]
 row 2: doc_ids = [2,2,2,2,2] 
 row 3: doc_ids = [2,2,2,2,2]   ← PAD row inherits doc 2's ids
                                                                                                                                                                                                                                 num_docs still 3   (because there are 3 unique doc_ids) so weights are set to 
 weights = [w0, w1, w2]   shape (3,)

(if doc_ids are absent, then num_docs is incremented correctly)

Then when we try to zero the weights of the padding row, we actually zero the last doc:
  worker: stream.weights.data[-pad_count:] = 0
 → weights = [1, 1, 0]

                                                                                                                                                           
New behaviour (fixed)                                                                                                                                                                                                           
                                                                                                                                                                                                                                  
pad_indices applied AND pad rows' doc_ids remapped to a synthetic id:                                                                                                                                                           

 row 0: doc_ids = [0,0,0,0,0]
 row 1: doc_ids = [1,1,1,1,1]
 row 2: doc_ids = [2,2,2,2,2]   ← real last doc
 row 3: doc_ids = [3,3,3,3,3]   ← synthetic doc for pad

  num_docs = 4,  weight_pad_count = 1
  weights = [w0, w1, w2, w3]   shape (4,)

  worker: stream.weights.data[-weight_pad_count:] = 0
        → weights = [1, 1, 1, 0]

  training lookup w[doc_id]:
    row 0 → w0 = 1  ✓
    row 1 → w1 = 1  ✓
    row 2 → w2 = 1  ✓ preserved
    row 3 → w3 = 0  ✓ pad silenced

(and this matches the behaviour when doc_ids is absent)

davidoj:
QOL: Magic now saves doc_ids.pt along with scores.pt. It's a helpful utility to map each score back to the indices of the original dataset. It's not strictly required: you can replay the shuffle/chunking of the original dataset carrying the indices through, and use this to map each score back to its original document. This seems a bit annoying and possibly prone to error, though, and saving the canonical mapping is a small overhead for a improved quality of life.

BUGFIX: Previously, if we ran magic per-document with packing, scores would be ordered the same way as the original dataset but without packing scores would be ordered as they were *after* shuffling the dataset. Now on both paths scores will be ordered the same way as the docs in the original dataset.

Tested equality of scores for per-doc unpacked vs aggregating per-token unpacked and per-doc packed vs aggregating per-token packed

davidoj:
QOL (?): Previously, loading datasets from the huggingface_hub respected the split argument, but loading local json/csv files ignored it. Changed to use `load_dataset("csv", data_files=..., split=...)` so that the split argument is respected. If we call dataset.load_from_disk directly we still don't support slicing and concatenation, and now raise an informative error if someone tries to pass a slice or concatenate splits on this path (instead of just raising a keyerror).

davidoj:
Added per-token validation plus a test that per-token validation produces the same scores as per-doc validation, when appropriately grouped.

docs/magic.rst
```diff
@@ -27,6 +27,45 @@ Usage
        --query.split "train[:8]" \
        --model EleutherAI/pythia-14m
 
+Output files
```
luciaquirke, line 30: 💯

luciaquirke (approved):
Fantastic, LGTM

PR #243, opened 2026-04-27
Compressed EKFAC index

jammastergirish:
**Draft — pre-PR review.**

## Summary

Adds `bergson compressed_ekfac` — a CLI subcommand + `compressed_ekfac_pipeline` orchestrator that produces a per-example (or per-token) gradient index where each row is `vec(P · H^{-1/2} · g)`. EKFAC preconditioning is baked in at build time, double-sided random projection applied after preconditioning. The index is queryable by plain dot product as the recall phase of two-stage retrieval; the existing per-example path supplies the precision phase.

Three first-class Hessian-approximation variants now share a unified `Preconditioner` protocol: `autocorrelation` (the existing trackstar dump), `kfac`, `ekfac`. Variant is auto-detected from the on-disk layout — no new config field. Builder + score paths route through `bergson.preconditioners.load_preconditioner` instead of hardcoding `get_trackstar_preconditioner`.

## How to review

The full design + alternatives considered + empirical findings + open questions live in [`COMPRESSED_EKFAC_PLAN.md`](https://github.com/jammastergirish/bergson/blob/feature/compressed-ekfac/COMPRESSED_EKFAC_PLAN.md). Suggested reading order:

1. §1 (what shipped + done bar)
2. §6 (empirical retrieval validation — the `projection_dim` floor finding is the meatiest deviation from plan)
3. §9 (questions where your call could redirect work)
4. §10 (commit log)
5. then the diff

## Headline empirical result (§6.4)

| Setting | Compressed-only recall@10 | Two-stage recall@10 (K1=20) | Spearman |
|---|---|---|---|
| pythia-160m, p=128, n_train=200, n_query=20 | 35.5 % | **48.5 %** | 0.397 |

108× compression, 145× faster Stage-1 lookup. Two-stage adds +13 pp recall over compressed-only.

## Open questions (full list in §9)

- **Math**: rotate-scale-rotate parity-tested vs `EkfacApplicator` at `atol=1e-5`, but if that itself is experimental rather than reference-trusted, a second reference (Kronfluence?) would let us triangulate. KFAC `Λ = Λ_G ⊗ Λ_A` orientation OK? Damping scheme for KFAC's outer-product Λ — same as EKFAC or different?
- **Naming**: rename `get_trackstar_preconditioner` in this PR or defer?
- **Score-time**: route compressed indices through `create_scorer` with a "pre-preconditioned" flag, or stay on the notebook dot-product path?
- **Bias + EKFAC**: refuse for MVP — OK?
- **pythia-1.4b stretch**: predicted to need `p ≥ 256` and ~50 GB factor budget per A40. Acceptable, or switch projection scheme?

## Test status

- Cluster `pytest tests/` at HEAD: **179 passed, 13 same pre-existing failures, 4 skipped, zero regressions** vs the pre-refactor baseline.
- 17 new tests added across the branch (12 unit-level for the new Preconditioner protocol + KFAC/EKFAC parity, 5 integration-level for `compressed_ekfac` end-to-end).
- Pre-existing failures unchanged, all unrelated (bitsandbytes dep, A40 numerical-determinism cache, `torch.optim.Muon` not in torch 2.6, `test_truncation` regex drift).

## Test plan

- [ ] Read `COMPRESSED_EKFAC_PLAN.md` end-to-end
- [ ] Reproduce §6.4 numbers via `scripts/validate_compressed_ekfac.py`
- [ ] Open `notebooks/compressed_ekfac_two_stage.ipynb` (outputs are embedded, renders on GitHub)
- [ ] Run `pytest tests/` locally; see §10 for expected count + pre-existing failures unrelated to this work


bergson/preconditioners.py
```diff
@@ -0,0 +1,361 @@
...
+        return cls(
+            q_a=q_a,
+            q_s=q_s,
+            lam=lam,
+            power=power,
+            lambda_damp_factor=lambda_damp_factor,
+        )
+
+
+def _detect_variant(path: Path) -> str:
```
luciaquirke, line 254: brittle. we should probably save the preconditioner mode as a class HessianMode(str, Enum)

examples/filter_data.py
```diff
@@ -394,7 +397,7 @@ def sum_(col):
         preconditioner_path = (
             args.query_dataset if args.query_dataset else args.index_dataset
         )
-        h_inv = get_trackstar_preconditioner(
+        h_inv = get_autocorrelation_preconditioner(
```
luciaquirke, line 400: lovely

bergson/process_grads.py
```diff
@@ -161,6 +172,13 @@ def get_trackstar_preconditioner(
     }
 
 
+# Backwards-compatible alias for the original "trackstar"-flavoured name.
+# Prefer ``get_autocorrelation_preconditioner`` in new code — see §2.1 of
+# COMPRESSED_EKFAC_PLAN.md for the naming rationale (autocorrelation is one
+# of three sibling Hessian-approximation types: autocorrelation / kfac / ekfac).
+get_trackstar_preconditioner = get_autocorrelation_preconditioner
```
jammastergirish, line 179: I actually asked for some of this as thought it may be necessary. To discuss!

luciaquirke:
We want to replace the "pipelined" type file like trackstar.py with a Pipeline runner that runs a series of commands specified in a yaml file (like bergson magic examples/magic/gpt_tiny.yaml, but with magic being a key inside the file). 

It will be more efficient to add the pipeline feature before we add the EK-FAC recall-optimized pipeline, so for this chunk of work I would do a Hessian approximation unification PR using the name Hessian (not Preconditioner), and then another PR to load the various Hessians and apply them during index build.

Then we could discuss the pipeline task, and the EK-FAC pipeline build task as future PRs.

PR #245, opened 2026-04-28
Unifies Hessian computation; renames preconditioner → hessian 

jammastergirish:
Unifies all Hessian estimation methods under one umbrella. The old `Preconditioners` CLI subcommand now just one `method` choice on the existing `Hessian` subcommand, and naming across the codebase follows: "preconditioner" → "hessian".

  **What changed:**                                       
   
  1. **`bergson preconditioners` subcommand removed;      
  folded into `bergson hessian` (1588e35).**
     - `HessianConfig.method` gains `"autocorrelation"`   
  and makes it the new default. `bergson hessian --method 
  autocorrelation` does what `bergson preconditioners`
  used to do; `--method kfac|tkfac|shampoo` are unchanged.
     - **Breaking:** scripts/configs invoking `bergson
  preconditioners ...` must move to `bergson hessian      
  --method autocorrelation ...`.
                                                          
  2. **Codebase-wide rename: `preconditioner*` →          
  `hessian*` (and `precond` shorthand → `hess`) 
  (546019d).** Verb forms (`preconditioning`,             
  `preconditioned`) preserved as descriptive math
  language. Public API renames worth flagging:
     - `GradientProcessor(preconditioners=…,
  preconditioners_eigen=…)` → `hessians=`,
  `hessians_eigen=`
     - `GradientProcessor.load(skip_preconditioners=…)` →
  `skip_hessians=`                                        
     - `bergson.mix_preconditioners` →
  `bergson.mix_hessians`                                  
     - Module renames: `bergson.process_preconditioners` →
   `bergson.process_hessians`; `bergson.collector.dist_pre
  conditioners_gradient_collector` →
  `bergson.collector.dist_hessians_gradient_collector`;   
  `examples.semantic.preconditioners` →    
  `examples.semantic.hessians`.

  3. **`GradientProcessor.load()` is backward compatible  
  `hessians_eigen.pth` are absent it falls back to 
  `preconditioners.pth` / `preconditioners_eigen.pth`.    
  Also defensively migrates any legacy `preconditioner*`
  keys appearing in `processor_config.yaml`.


luciaquirke:
Looking awesome. Two things:

Where you insert backwards compatibility could you please add a
```# TODO Lucia Quirke: remove on the 28 October 2026```

Comment so I can sweep through and remove it at some point when people are safely migrated.


Second,  there's a user-facing flag in config.py formerly called preconditioner_path that's used here

    # Load preconditioners on device one-by-one for memory efficiency
    preconditioners = GradientProcessor.load(
        Path(preconditioner_path),
        map_location="cpu",
    ).preconditioners

I think it should be called "processor_path" because it's actually a path to a GradientProcessor. And its docstring was """Path to a precomputed preconditioner.""" I think it should be """Path to a precomputed gradient processor. Set to apply a Hessian approximation."""

I think this is 95% ready

**Comment by luciaquirke** [comment]
@LouisYRYJ @SrGonao @norabelrose we're interested in making this refactor that makes a small change that nonetheless touches many files to move towards unified handling for our various Hessian approximations. But if this change seems too large or risky, the first commit (unifying the command line tools and nothing else) could be merged now along with the `preconditioner_path` -> `processor_path` rename, and we could leave the larger naming refactor for later (post paper submissions). Please shout out if you want this!  Otherwise we'll make the call in around 24 hours.

File: benchmarks/benchmark_factors.py [hunk 1]
```diff
@@ -34,7 +34,7 @@ def _model_device(model: torch.nn.Module) -> torch.device:
     return next(model.parameters()).device
 
 
-BERGSON_FACTOR_TYPES = {"normalizer", "preconditioner", "kfac", "ekfac"}
+BERGSON_FACTOR_TYPES = {"normalizer", "hessian", "kfac", "ekfac"}
```
LouisYRYJ: here I think we should say something like "auto_correlation_hessian" or maybe "ac_hessian", probably there are better names out there

LouisYRYJ:
Thanks a lot Girish, this has been laying around for quite some time and I am happy to see this being picked up.
The changes look good to me! I have a few comments, not very loadbearing, and from my side we are good to merge.

- In general, I would like to avoid calling the autocorrelation projected Hessian the "hessian", using it as a variable name is fine, but I left an example where I think it may be a bit misleading? 

(Importantly, we are at no point computing the actual true Hessian in this library, but always approximations thereof. This makes it generally difficult to be precise when naming things and talking about them, but I have found no good solution for this and also defaulted to calling some functions/files "hessian_sth_sth" for the sake of brevity even though that is technically not correct.)

- **Out of scope for this PR**, but one place where we have some redundancy: GradientCollectorWithDistributedHessians is using a different distributed logic than for the Hessian approximations I have been using. The first shards tensors according to modules, while the latter shards tensors themselves (using the 0th dim to split them across ranks). The second right now doesn't handle the case where tensor.shape[0] is not divisible by world_size gracefully, but in general is fairer and ensures that memory is distributed as fairly as possible (which in this case is pretty important). I would definitely like to handle the "tensor.shape[0] is not divisible by world_size" case in future (presumably here https://github.com/jammastergirish/bergson/blob/c267fcb394ca40814c5d87714c9d0ba6bc10fbd8/bergson/hessians/sharded_computation.py)

- Small nit: "Hessian" is a noun, so in docs we should capitalize it

Looking forward to your future contributions :)

jammastergirish:
Thank you so much for this @LouisYRYJ.

I'll discuss with @luciaquirke the language to use. Agree that `hessian` may be too broad.

Let's deal with that redundancy in another PR.

And thanks on the proper noun capitaliztion!

benchmarks/benchmark_factors.py
luciaquirke: `autocorrelation` make sense to me since the others don't have the _hessian suffix

luciaquirke:
Yeah this is somewhat gnarly but it would be really good to find anywhere where we did a "preconditioner" -> "hessian" rename where it would be more appropriate to do "preconditioner" -> "autocorrelation" (like if we're specifically referring to the autocorrelation approximation). I will sweep through and try to do this now.

Re: calling our Hessian approximations "hessian" rather than "hessian_approx", it's definitely not ideal but probably worth it for the conciseness gains.

benchmarks/benchmark_factors.py
```diff
@@ -57,7 +57,7 @@ class RunConfig:
     """Method to benchmark: bergson, kronfluence, or dattri."""
 
     factor_type: str = field(positional=True)
-    """Factor type. bergson: normalizer/preconditioner/kfac/ekfac.
+    """Factor type. bergson: normalizer/hessian/kfac/ekfac.
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

benchmarks/benchmark_factors.py
```diff
@@ -144,14 +144,14 @@ def _get_factor_run_path(
 # ---------------------------------------------------------------------------
 
 
-def _run_bergson_preconditioner(
+def _run_bergson_hessian(
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

benchmarks/benchmark_factors.py
```diff
@@ -144,14 +144,14 @@ def _get_factor_run_path(
...
     run_cfg: RunConfig,
     spec,
     train_tokens: int,
     run_path: Path,
     ds,
 ) -> tuple[float, float]:
-    """Run bergson preconditioner (P^T@P + eigendecomp).
+    """Run bergson hessian (P^T@P + eigendecomp).
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

benchmarks/plot_factor_benchmark.py
```diff
@@ -32,7 +32,7 @@
 # Hatching per factor type for visual distinction within a method
 FACTOR_HATCHES: dict[str, str] = {
     "normalizer": "",
-    "preconditioner": "//",
+    "hessian": "//",
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

bergson/collector/dist_hessians_gradient_collector.py
```diff
@@ -10,17 +10,17 @@
 
 from bergson.collector.collector import HookCollectorBase
 from bergson.config import IndexConfig
-from bergson.process_preconditioners import process_preconditioners
+from bergson.process_hessians import process_hessians
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

bergson/collector/dist_hessians_gradient_collector.py
```diff
@@ -10,17 +10,17 @@
...
 from bergson.utils.utils import get_gradient_dtype
 
 
 @dataclass(kw_only=True)
-class GradientCollectorWithDistributedPreconditioners(HookCollectorBase):
+class GradientCollectorWithDistributedHessians(HookCollectorBase):
```
luciaquirke: let's rename "preconditioner" -> "autocorrelation" here

bergson/collector/gradient_collectors.py
```diff
@@ -91,16 +91,16 @@ def setup(self) -> None:
 
     @HookCollectorBase.split_attention_heads
     def backward_hook(self, module: nn.Module, g: Float[Tensor, "N S O"]):
-        """Compute per-sample gradient, accumulate preconditioner, and store."""
+        """Compute per-sample gradient, accumulate hessian, and store."""
```
luciaquirke: "autocorrelation matrix" in this file

bergson/__init__.py
```diff
@@ -54,5 +54,5 @@
     "Scorer",
     "ScoreConfig",
     "QueryConfig",
-    "mix_preconditioners",
+    "mix_hessians",
```
luciaquirke: This public API is unique in that it will always be specific to the autocorrelation matrix approximation, so let's call it mix_autocorrelation_matrices

bergson/__main__.py
```diff
@@ -77,8 +77,15 @@ class Hessian(Serializable):
...
+
         validate_run_path(self.index_cfg)
-        approximate_hessians(self.index_cfg, self.hessian_cfg)
+
+        if self.hessian_cfg.method == "autocorrelation":
+            self.skip_index = True
+            self.skip_hessians = False
+            build(self.index_cfg, PreprocessConfig())
+        else:
+            approximate_hessians(self.index_cfg, self.hessian_cfg)
```
luciaquirke, line 88: Nice 🔥

docs/cli.rst
```diff
@@ -2,7 +2,7 @@ Command Line Interface
 ======================
 
 Bergson's post-hoc attribution exposes three building block commands — ``build``, ``reduce``, and
-``score`` — plus supporting commands for querying, preconditioner computation, and
+``score`` — plus supporting commands for querying, hessian computation, and
 end-to-end pipelines.
```
luciaquirke, line 6: Nice

docs/pipeline.rst
```diff
@@ -21,10 +21,10 @@ The difference between them is **what they do with the collected gradients**:
 - ``reduce`` **aggregates** all gradients from a dataset into a single vector and writes it to an on-disk file.
 - ``score`` **computes similarity scores** by comparing gradients from one dataset against a pre-built query.
 
-The supporting commands handle preconditioner computation and end-to-end pipelines:
-
-- ``hessian`` computes Hessian approximations (KFAC, TKFAC, Shampoo) stored as sharded covariance matrices.
-- ``preconditioners`` fits normalizers and preconditioners without collecting gradients.
+The supporting command ``hessian`` computes Hessian approximations
+(``autocorrelation`` — the gradient second-moment / Adam-style estimator and the
```
luciaquirke: Remove reference to "Adam-style estimator", this is a legacy thing

bergson/collector/gradient_collectors.py
jammastergirish: "accumulate autocorrelation matrix" or "autocorrelation matrix"?

luciaquirke: accumulate for sure, the accumulation is the part where we sum all the outer product matrices to compute an autocorrelation matrix

jammastergirish: Will make that very quick change shortly.

PR #246, opened 2026-04-28
ELE-11: Add YAML-mediated search pipeline to Bergson

jammastergirish:
See notes in [Linear](https://linear.app/eleutherai-interpretability/issue/ELE-11/add-yaml-mediated-search-pipeline-to-bergson) on the test, which seemed to expose a bug deeper in the library. (I've now dealt with this bug [here](https://github.com/EleutherAI/bergson/pull/252).

Adds a generic `bergson pipeline <file.yaml>` command that runs a sequence of existing bergson subcommands defined in a YAML file. This unblocks pipelines (`build_EKFAC.yaml`, `query_EKFAC.yaml`, …) without writing a new Python entry point per pipeline.

- **New module** — `bergson/yaml_pipeline.py`: `parse_pipeline()` reads the YAML once, validates its shape, and hydrates each step into the matching command dataclass via `simple_parsing`'s `Serializable.from_dict`. Defaults from the existing dataclasses fill in any unspecified fields automatically. All steps are parsed up-front so a config error in step N fails before step 1 burns compute. `run_pipeline()` parses then executes in order.
- **`bergson/__main__.py`**: adds a third branch to `main()` for pipeline mode alongside the existing single-command-from-file and CLI-flag modes. Hoists the `{name: class}` registry out of the existing branch so both branches share it. 
- **`pyproject.toml`**: declares `pyyaml` as a direct dep. (It was already required transitively by `Serializable.load` for the existing single-command YAML flow; this just documents that.)
- **Example** — `examples/pipelines/hessian_then_build.yaml`: minimal hessian-then-build pipeline.

## YAML format

A pipeline is a **list of single-key mappings**. Each entry's key is a registered command name (case-insensitive) and its value is the same config tree that command accepts as a single-command YAML so each step is fully self-contained.

```yaml
- hessian:
    hessian_cfg:
      method: kfac
    index_cfg:
      run_path: runs/example
      model: gpt2
- build:
    index_cfg:
      run_path: runs/example
      model: gpt2
    preprocess_cfg: {}
```

A list (not a mapping) is required because YAML mappings can't preserve order or duplicates, and pipelines need both.

## What's deferred

- **Shared values across steps** (e.g. one `run_path` injected into every step) are not supported — each step must currently repeat its config. Easy to layer on later via a top-level `defaults:` block; should we do this, @luciaquirke?
- **`Test_Model_Configuration` can't be a pipeline step** because `DiagnoseConfig` isn't `Serializable`. This is a pre-existing limitation — `bergson test_model_configuration <yaml>` doesn't work today either — and out of scope here.

## Follow-up needed for `build_EKFAC.yaml` and `query_EKFAC.yaml`

- **`build_EKFAC.yaml`** would work today by wrapping the existing `hessian` command.
- **`query_EKFAC.yaml`** is partially blocked: step 3 of the EKFAC flow (\"apply inverse Hessian to the mean query gradient\") is not currently exposed as a top-level CLI command — it lives as `apply_worker` inside `bergson/hessians/pipeline.py`. A small follow-up PR will lift it into an `apply_hessian` command so the full query pipeline can be expressed as YAML.

(@luciaquirke I had a go at this with Claude [here](https://github.com/EleutherAI/bergson/pull/247) but closed that PR so we can focus on this one. Not looked at that yet.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,16 @@ You can also aggregate your query dataset into a single mean or sum gradient as 
 bergson build <output_path> --model <model_name> --dataset <dataset_name> --aggregation mean --unit_normalize --preconditioner_path <path_to_preconditioner>
 ```
 
+## Run a Multi-Step Pipeline
+
+Many workflows chain several Bergson commands together. Rather than running each command separately, you can express the whole sequence as a YAML file and run it with a single command:
+
+```bash
+bergson pipeline <path_to_yaml>
+```
+
+The YAML is a list of single-key entries, one per step, each holding that command's full config. See [`examples/pipelines/hessian_then_build.yaml`](examples/pipelines/hessian_then_build.yaml) for a runnable Hessian → build example.
+
 ## Query an On-Disk Gradient Index
 
 We provide a query Attributor which supports unit normalized gradients and KNN search out of the box. Access it via CLI with
--- a/bergson/__main__.py
+++ b/bergson/__main__.py
@@ -24,6 +24,7 @@ from .query.query_index import query
 from .score.score import score_dataset
 from .trackstar import trackstar
 from .utils.worker_utils import validate_run_path
+from .yaml_pipeline import run_pipeline
 
 
 @dataclass
@@ -199,18 +200,38 @@ class Main:
 
 
 def main():
-    """Parse CLI arguments and dispatch to the selected subcommand."""
-    # Check to see if the user is passing in a yaml or json config file directly
+    """Parse CLI arguments and dispatch to the selected subcommand.
+
+    Three input shapes are supported:
+      1. `bergson pipeline <file.yaml>`        — multi-step pipeline mode
+      2. `bergson <command> <file.yaml|json>`  — single-command config-file mode
+      3. `bergson <command> --flag value ...`  — single-command CLI-flag mode
+    """
     args = sys.argv[1:]
+
+    # Build the {command_name: command_class} lookup once; both the pipeline
+    # branch and the single-command branch use it to resolve the user's verb.
+    command_classes = get_args(Main.__dataclass_fields__["command"].type)
+    command_registry = {cls.__name__.lower(): cls for cls in command_classes}
+
+    # Branch 1 — Pipeline mode: a YAML listing several commands to run in
+    # sequence. Delegates parsing + execution to `run_pipeline`.
+    if len(args) == 2 and args[0].lower() == "pipeline" and os.path.isfile(args[-1]):
+        run_pipeline(args[1], command_registry)
+        return
+
+    # Branch 2 — Single-command config-file mode: the user passes a command
+    # name and a path to a YAML or JSON file. `Serializable.load` sniffs the
+    # file extension and hydrates the matching dataclass.
     if len(args) == 2 and os.path.isfile(args[-1]):
         cmd_str, config_path = args
-
-        args = get_args(Main.__dataclass_fields__["command"].type)
-        names = {cls.__name__.lower(): cls for cls in args}
         try:
-            cmd_cls = names[cmd_str.lower()]
+            cmd_cls = command_registry[cmd_str.lower()]
         except KeyError:
-            print(f"Invalid command '{cmd_str}'. Valid commands are: {list(names)}")
+            print(
+                f"Invalid command '{cmd_str}'. "
+                f"Valid commands are: {list(command_registry)}"
+            )
             sys.exit(1)
 
         try:
@@ -218,6 +239,7 @@ def main():
         except RuntimeError as e:
             print(f"Failed to load config file {config_path}: {e}")
             sys.exit(1)
+    # Branch 3 — CLI-flag mode: standard argparse-style flag parsing.
     else:
         parser = ArgumentParser(conflict_resolution=ConflictResolution.EXPLICIT)
         parser.add_arguments(Main, dest="prog")
--- a/bergson/__main__.py
+++ b/bergson/__main__.py
@@ -203,9 +203,10 @@ def main():
     """Parse CLI arguments and dispatch to the selected subcommand.
 
     Three input shapes are supported:
-      1. `bergson pipeline <file.yaml>`        — multi-step pipeline mode
-      2. `bergson <command> <file.yaml|json>`  — single-command config-file mode
-      3. `bergson <command> --flag value ...`  — single-command CLI-flag mode
+      `bergson <command> --flag value ...`  — single-command CLI-flag mode
+      `bergson pipeline <file.yaml>`        — multi-step pipeline mode
+      `bergson <command> <file.yaml|json>`  — single-command config-file mode
+
     """
     args = sys.argv[1:]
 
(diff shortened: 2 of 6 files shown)


luciaquirke:
> pyproject.toml: declares pyyaml as a direct dep. (It was already required transitively by Serializable.load for the existing single-command YAML flow; this just documents that.)

Documentation is not why we list direct dependencies

luciaquirke:
> Shared values across steps (e.g. one run_path injected into every step) are not supported — each step must currently repeat its config. Easy to layer on later via a top-level defaults: block; should we do this, @luciaquirke?

No, sounds too complex

jammastergirish:
> > pyproject.toml: declares pyyaml as a direct dep. (It was already required transitively by Serializable.load for the existing single-command YAML flow; this just documents that.)
> 
> Documentation is not why we list direct dependencies

It's not just documentation. What if, for example, the first library drops the one that's not explicitly named and replaces it with another?

luciaquirke:
I can't see that it's not passing typecheck yet but the PR looks amazing! Let me know when it's ready for review

bergson/__main__.py
```diff
@@ -199,25 +200,46 @@ def execute(self):
 
 
 def main():
-    """Parse CLI arguments and dispatch to the selected subcommand."""
-    # Check to see if the user is passing in a yaml or json config file directly
+    """Parse CLI arguments and dispatch to the selected subcommand.
+
+    Three input shapes are supported:
+      1. `bergson pipeline <file.yaml>`        — multi-step pipeline mode
```
luciaquirke: Order from simplest to most complex (reverse order)

jammastergirish:
> I can't see that it's not passing typecheck yet but the PR looks amazing! Let me know when it's ready for review

Thanks! 

I think it's ready notwithstanding the bug mentioned in Linear: https://linear.app/eleutherai-interpretability/issue/ELE-11/add-yaml-mediated-search-pipeline-to-bergson#comment-c68c7bfc

bergson/__main__.py
jammastergirish: Good idea

bergson/__main__.py
```diff
@@ -199,25 +200,46 @@ def execute(self):
...
+      2. `bergson <command> <file.yaml|json>`  — single-command config-file mode
+      3. `bergson <command> --flag value ...`  — single-command CLI-flag mode
+    """
     args = sys.argv[1:]
+
+    # Build the {command_name: command_class} lookup once; both the pipeline
+    # branch and the single-command branch use it to resolve the user's verb.
+    command_classes = get_args(Main.__dataclass_fields__["command"].type)
+    command_registry = {cls.__name__.lower(): cls for cls in command_classes}
```
luciaquirke, line 216: I think `.lower()` could lead to name clashes?

bergson/__main__.py
```diff
@@ -199,25 +200,46 @@ def execute(self):
...
         try:
-            cmd_cls = names[cmd_str.lower()]
+            cmd_cls = command_registry[cmd_str.lower()]
         except KeyError:
-            print(f"Invalid command '{cmd_str}'. Valid commands are: {list(names)}")
+            print(
+                f"Invalid command '{cmd_str}'. "
+                f"Valid commands are: {list(command_registry)}"
+            )
             sys.exit(1)
```
luciaquirke: nit: I know this is existing code but why not `raise e`?

bergson/__main__.py
```diff
@@ -199,25 +200,46 @@ def execute(self):
...
+      3. `bergson <command> --flag value ...`  — single-command CLI-flag mode
+    """
     args = sys.argv[1:]
+
+    # Build the {command_name: command_class} lookup once; both the pipeline
+    # branch and the single-command branch use it to resolve the user's verb.
+    command_classes = get_args(Main.__dataclass_fields__["command"].type)
+    command_registry = {cls.__name__.lower(): cls for cls in command_classes}
+
+    # Branch 1 — Pipeline mode: a YAML listing several commands to run in
```
luciaquirke: If you add numbered comments then if anyone adds a new branch they will have to update all the numbers so I recommend avoiding that

bergson/__main__.py
```diff
@@ -199,25 +200,46 @@ def execute(self):
...
+    # sequence. Delegates parsing + execution to `run_pipeline`.
+    if len(args) == 2 and args[0].lower() == "pipeline" and os.path.isfile(args[-1]):
+        run_pipeline(args[1], command_registry)
+        return
+
+    # Branch 2 — Single-command config-file mode: the user passes a command
+    # name and a path to a YAML or JSON file. `Serializable.load` sniffs the
```
luciaquirke: Serializable.load internals description is TMI

bergson/__main__.py
jammastergirish: Agree

bergson/__main__.py
jammastergirish: Good idea

bergson/__main__.py
jammastergirish: Good point

luciaquirke:
This is looking really fabulous!! I added what amounts to style nits. With a green build and those addressed, LGTM.

examples/pipelines/hessian_then_build.yaml
```diff
@@ -0,0 +1,23 @@
+# Each list entry is a single registered command (see `bergson --help`).
+# Each step is fully self-contained: it carries every sub-config the
+# matching command would need on its own. There is intentionally no
+# top-level shared state — repeat values like `run_path` in every step.
```
luciaquirke: Extraneous comment, remove

examples/pipelines/hessian_then_build.yaml
```diff
@@ -0,0 +1,23 @@
...
+        chunk_length: 1024
+
+- build:
+    index_cfg:
+      run_path: runs/autocorrelation_and_index
+      model: gpt2
+      token_batch_size: 1024
+      data:
+        chunk_length: 1024
+    preprocess_cfg: {}
```
luciaquirke, line 24: Noting that this yaml file doesn't do what you would expect because we don't yet support building a Hessian over compressed gradients, so the Hessian and the (compressed by default) index here would be incompatible

bergson/__main__.py
jammastergirish: Actually, I think not, as this would only be an issue if we had conflicting names ourselves, which we wouldn't (unless we coded badly). Removing lower forces the user to always get the case right in the command line, which is not always going to happen. Do you agree @luciaquirke?

luciaquirke: oh right! Yeah I agree with you, I misunderstood this code

luciaquirke (approved):
Looking great! Ideally we'd hold off on merging the hessian_then_build.yaml until we have EK-FAC support in `score`, at which point we'd merge it as a `hessian_then_score.yaml`. Everything else looks ready to merge.

README.md
```diff
@@ -103,6 +103,16 @@ You can also aggregate your query dataset into a single mean or sum gradient as
 bergson build <output_path> --model <model_name> --dataset <dataset_name> --aggregation mean --unit_normalize --preconditioner_path <path_to_preconditioner>
 ```
 
+## Run a Multi-Step Pipeline
```
luciaquirke, line 106: Love this!!!

luciaquirke (approved):

PR #253, opened 2026-04-29
Add build → score pipeline example

jammastergirish:
## Summary

Adds a runnable two-step pipeline example demonstrating the no-compression scoring path:

```bash
bergson pipeline examples/pipelines/build_then_score.yaml
```

Step 1 (`build`) creates a one-gradient on-disk query index with `projection_dim: 0` and `aggregation: mean`. Step 2 (`score`) loads that query into memory and dot-products it against each item in a small slice of `NeelNanda/pile-10k`, producing per-item influence scores. Until now, this required two CLI invocations — this single YAML now drives the whole flow through `bergson pipeline`.

## Why

1. **Showcase the YAML pipeline functionality** introduced in #246 with a real, useful example.
2. **Demonstrate a feature that previously required multiple commands** (`bergson build` + `bergson score`) — now reducible to a single `bergson pipeline` invocation.
3. **Lay groundwork for EK-FAC support in `score`** — having an end-to-end no-compression scoring pipeline in YAML form is a prerequisite for then layering Hessian application onto the uncompressed gradients (Lewis's worker has the existing reference for that), without entangling Hessian work with the rest of the codebase.

## Files

- `examples/pipelines/build_then_score.yaml` — the example pipeline (small model: `gpt2`, small dataset: `NeelNanda/pile-10k` with `train[:20]` for the query and `train[:100]` for scoring, `chunk_length: 1024`).
- `README.md` — extends the "Run a Multi-Step Pipeline" section introduced in #246 with a one-sentence reference to `build_then_score.yaml` as a second example covering the no-compression scoring path. (The section itself, the description of `bergson pipeline`, and the link to `hessian_then_build.yaml` all live in #246.)
- `tests/test_yaml_pipeline.py` — `test_build_then_score_example_parses` asserts the shipped YAML hydrates into the right `Build`/`Score` commands with the expected configs (parse-only, no GPU needed).

## Configuration notes

- **`skip_preconditioners: true`** is set on both build and score steps. With `projection_dim: 0` the per-token gradient `P` retains the full flattened module-gradient dimension; the default per-module Adafactor preconditioner update `P.mT @ P` would otherwise materialize a `(full_grad_dim × full_grad_dim)` matrix — multi-TB even for GPT-2's `c_attn` — and OOM immediately. The no-compression scoring path uses raw dot products of full gradients, so preconditioners aren't needed here. See issue #255 for a suggested follow-up to make this implicit.
- **`unit_normalize` and `--preconditioner_path`** from the CLI command in the Linear task aren't replicated here. Those belong to the compressed/preconditioned path. This example is deliberately the no-compression demo.

## Dependencies

- **Stacked on #246** (`add_yaml-mediated_search_pipeline`) — this branch sits on top of the YAML pipeline runner introduced there. The PR base reflects that.
- **Requires #252** (`Add length column to tokenize_and_chunk output`) to run successfully. Without it the build step crashes at `ValueError: Column 'length' doesn't exist.` because `tokenize_and_chunk` (triggered by `chunk_length: 1024`) doesn't add the column expected by the build/score workers. Please merge #252 before this.
- **Requires #251** (`torch>=2.4`) — the CLI fails to import on torch < 2.4 due to FSDP2 `fully_shard`. Not blocking for review but blocks running the example on stale environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Many workflows chain several Bergson commands together. Rather than running each
 bergson pipeline <path_to_yaml>
 ```
 
-The YAML is a list of single-key entries, one per step, each holding that command's full config. See [`examples/pipelines/hessian_then_build.yaml`](examples/pipelines/hessian_then_build.yaml) for a runnable Hessian → build example.
+The YAML is a list of single-key entries, one per step, each holding that command's full config. See [`examples/pipelines/hessian_then_build.yaml`](examples/pipelines/hessian_then_build.yaml) for a runnable Hessian → build example, or [`examples/pipelines/build_then_score.yaml`](examples/pipelines/build_then_score.yaml) for a build → score example using the no-compression scoring path (`projection_dim: 0`).
 
 ## Query an On-Disk Gradient Index
 
--- a/examples/pipelines/build_then_score.yaml
+++ b/examples/pipelines/build_then_score.yaml
@@ -1,0 +1,38 @@
+# Build a one-gradient query index, then score a dataset against it.
+#
+# This pipeline demonstrates the no-compression scoring path: gradients
+# are kept at full model dimensionality (`projection_dim: 0`) and
+# influence is computed as a raw dot product between the query gradient
+# and each training gradient.
+#
+# Run with:
+#     bergson pipeline examples/pipelines/build_then_score.yaml
+
+- build:
+    index_cfg:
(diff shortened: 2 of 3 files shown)


luciaquirke (approved):
LGTM! We will add a version of this YAML with Hessian fitting + application at score and build time next :)

PR #258, opened 2026-04-30
Remove user-facing magic non-FSDP assert since we do use FSDP, just a different implementation

luciaquirke:
MAGIC now supports both SimpleFSDP and DDP. In the script we automatically don't try to use FSDP2 since we use FSDP. I think the FSDP flag should turn on SimpleFSDP in the MAGIC context and FSDP2 in other contexts, and this is what happens. so the assert seems unnecessary.


norabelrose (approved):
lgtm

PR #259, opened 2026-05-01
Refactor load_from_optimizer

luciaquirke:
**Changes**
- Use enum for optimizer format switching
- Claude-suggested bug fix: manually append the active adapter name in `load_from_optimizer`. Details:

Our function `extract_peft_target_modules` produces module names with the suffix (`...lora_A.default`).

`load_from_optimizer` uses `peft.get_peft_model_state_dict` to get module names. This function returns LoRA keys without the suffix that's like `...lora_A.weight`.

Then every module is filtered in the normalizer loading loop:

```
if target_modules is not None and module_name not in target_modules:                                               
      continue  
```

and `load_from_optimizer` raises a "no second moments" assertion. 

**New tests**
- Test the regression (PEFT with adapter-suffixed target_modules loads, stripped target_modules raises)
- Test a few details like bad/empty target_modules, unrecognised state entries, include_bias on Adam, and unit tests for the new format helpers.

### DIFF
--- a/bergson/utils/load_from_optimizer.py
+++ b/bergson/utils/load_from_optimizer.py
@@ -9,12 +9,12 @@ from bergson.gradients import AdafactorNormalizer, AdamNormalizer, Normalizer
 
 
 class OptimizerStateFormat(Enum):
-    """Optimizer state format for a single module - Adafactor for factored
-    optimizer states (e.g. 2D+ modules in Adafactor optimizers), and Adam
-    for unfactored states."""
+    """Optimizer state format for a single module - Adafactor-style factored
+    optimizer states (e.g. 2D+ modules in Adafactor optimizers), or Adam-style
+    unfactored states."""
 
-    ADAM = 1
-    ADAFACTOR = 2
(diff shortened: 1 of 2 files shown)


bergson/utils/load_from_optimizer.py
```diff
@@ -64,97 +133,47 @@ def load_from_optimizer(
...
     if isinstance(model, PeftModel):
         st = get_peft_model_state_dict(model)
         params_for_index = list(st.items())
+        # peft serializes LoRA keys without the active adapter name (e.g.
+        # ``...lora_A.weight``), but extract_peft_target_modules and the
+        # actual submodule paths include it (``...lora_A.default``). Append
+        # the adapter name so module_name lookups match target_modules.
+        adapters = list(model.peft_config.keys())
+        if len(adapters) == 1:
+            adapter_suffix = "." + adapters[0]
```
luciaquirke, line 154: @norabelrose this is the main thing I'm requesting review on, mostly FYI. It works in my LESS example script

PR #260, opened 2026-05-01
Add LESS example script

luciaquirke:
Adding this because we mention it in the paper, nice to show how it fits into the bergson world.

It uses TRAK-style full gradient projection so I added that using the TRAK library, they have a fast path for the projection using a custom kernel. I'm happy to split this feature out into a separate PR if that would make it easier to review, but it's basically TRAK-style projection changes in the `bergson` directory and the actual LESS changes in `examples/less`.


LouisYRYJ:
Can you say a bit more about this:
- Is this also a Rademacher projection but just faster? 
- If yes, should we (in future) just migrate to that?
- I thought LESS is an algorithm that we basically already supported?

luciaquirke:
> Can you say a bit more about this:
> 
> * Is this also a Rademacher projection but just faster?
> * If yes, should we (in future) just migrate to that?
> * I thought LESS is an algorithm that we basically already supported?

I thought we did but I went through the paper again and they use a projection over the full flattened model gradients, rather than our per-module two sided projection.  With this addition, we should have proper and correct support.

I haven't benchmarked their Rademacher projection against ours, and since it's a vector matrix mul rather than a matmul the perf characteristics might be meaningfully different in the per-module setting. We could check it out.

luciaquirke:
@norabelrose if you happen to know how our mat-mat rademacher projection compares to the custom kernel vec-mat one in the TRAK library we would be interested to know (not urgent)

PR #261, opened 2026-05-01
SOURCE implementation

LouisYRYJ:
This PR implements SOURCE from https://arxiv.org/abs/2405.12186

Steps:
1. Compute all objects that are needed
- [x]  EKFAC, or other approximators, for all L checkpoints
- [x] Compute query gradients at final checkpoint
- [x] Compute training grads at correct checkpoints (maybe we do this in run)
2. Do correct accumulation
- [x] Aggregate Hessians according to the predefined amount of stages
3. Put everything together E2E
- [x] Done


LouisYRYJ:
First minimal version done:
E2E is running, need to carefully validate steps

bergson/__main__.py
```diff
@@ -31,6 +32,35 @@
...
+
+    hessian_cfg: HessianConfig
+
+    approx_unrolling_cfg: ApproxUnrollingConfig
+
+    resume: bool = False
+    """If true, skip steps whose output directories already exist."""
+
+    def execute(self):
+        from bergson.approx_unrolling.pipeline import approx_unrolling_pipeline
```
luciaquirke: Imports at the top

bergson/hessians/hessian_approximations.py
```diff
@@ -226,6 +297,11 @@ def collect_hessians(
         cfg=index_cfg,
     )
 
-    computer.forward_backward = fwd_bwd_hessian_factory(index_cfg, hessian_cfg)
+    if (not ev_correction) and hessian_cfg.method == "foof":
```
luciaquirke: We should probably mention FOOF in the paper

bergson/collector/collector.py
```diff
@@ -850,6 +850,19 @@ def fwd_bwd_hessian(model, x: Tensor, y: Tensor, batch: dict):
     return fwd_bwd_hessian
 
 
+def fwd_only_factory(index_cfg: IndexConfig, hessian_cfg: HessianConfig) -> Callable:
```
luciaquirke: In my mind the point of the factory pattern is that you don't have to add a new factory for every feature, can we toggle between forward mode and otherwise in the hessian fwd bwd factory based on an arg in the Hessian config, e.g. method="foof", or is that too conceptually confusing. Whatever y'all decide is fine with me, just noticing 3 very similar factory methods here

bergson/approx_unrolling/approx_unrolling_math.py
```diff
@@ -0,0 +1,252 @@
...
+        x = eta_K * sigma
+        is_zero = x == 0
+        x_safe = x.masked_fill(is_zero, 1.0)
+        ratio = -torch.expm1(-x_safe) / x_safe
+        return eta_K * ratio.masked_fill(is_zero, 1.0)
+
+    return fn
+
+
+def apply_eigfn_to_query(
```
luciaquirke, line 91: I'm seeing a few words without underscores separating them (phase1 -> phase_1) and some initialisms without definitions (F_r, F_S). Either defining initialisms in the docstring or expanding them out so the names are interpretable seems pretty important to make this code maintainable.

bergson/approx_unrolling/approx_unrolling_math.py
```diff
@@ -0,0 +1,252 @@
...
+    ScoreConfig,
+)
+from bergson.data import load_scores
+from bergson.distributed import launch_distributed_run
+from bergson.hessians.apply_hessian import EkfacApplicator, EkfacConfig
+from bergson.score.score import score_dataset
+from bergson.utils.utils import get_device, get_device_index
+
+
+def compute_eta_K_per_segment(
```
luciaquirke: Continuing with the maintainability stuff, I think lr would more readable and consistent than eta. like

```
def sum_segment_step_lrs
```

or something

bergson/approx_unrolling/approx_unrolling_math.py
```diff
@@ -0,0 +1,252 @@
...
+    """Phase 1: build u_0, u_1, ..., u_{L-1} by walking F_S backwards.
+
+    u_{L-1} is the original query at <run>/query/.
+    u_{k-1} = F_S(segment_k) applied to u_k for k = L-1, L-2, ..., 1.
+    Outputs land at <run>/segment_{l}/u/ for l = 0 .. L-2.
+
+    Returns ``[u_0_path, u_1_path, ..., u_{L-1}_path]``.
+    """
+    base = Path(run_path)
+    L = len(eta_K_per_segment)
```
luciaquirke: num_segments

bergson/approx_unrolling/approx_unrolling_math.py
```diff
@@ -0,0 +1,252 @@
...
+        apply_eigfn_to_query(
+            src_grad_path=u_paths[k],
+            dst_grad_path=dst,
+            segment_dir=segment_dir,
+            eta_K=eta_K_per_segment[k],
+            n_seg=_load_n_seg(segment_dir),
+            fn_kind="f_s",
+            distributed=distributed,
+        )
+        u_paths[k - 1] = dst
```
luciaquirke: let's expand out the mysterious u. `training_state_paths`? or `unrolling_state_paths` maybe?

bergson/approx_unrolling/approx_unrolling_math.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    base = Path(run_path)
+    L = len(eta_K_per_segment)
+    u_paths: list[Path] = [Path("")] * L
+    u_paths[L - 1] = base / "query"
+
+    for k in range(L - 1, 0, -1):
+        segment_dir = base / f"segment_{k}" / method
+        dst = base / f"segment_{k - 1}" / "u"
+        apply_eigfn_to_query(
```
luciaquirke, line 160: I guess this is being applied to an unrolling state which is initially the query and later an unrolling state? Shall we call it apply_eigfn_to_unrolling_state or something? The MAGIC implementation has quite nice names for very similar operations. I think they have a TrainerState or BackwardState or something

bergson/hessians/hessian_approximations.py
LouisYRYJ: I cut it out for now, it's complicates the current pipeline by a lot and not that load bearing imo (we already have shampoo which is 100% more than what other libraries have)

bergson/collector/collector.py
LouisYRYJ: ageed, will need to rethink this pipeine and cut out foof for now

LouisYRYJ:
@luciaquirke Thanks, had a go and simplified a lot of stuff, hope this is more transparent now

PR #267, opened 2026-05-03
CUDA isolation + upfront VRAM usage check

luciaquirke:
I noticed GPU 0 had an extra .7 GB in VRAM usage per device. I've empirically observed this really adding up at scale.

dist.init_process_group("nccl", ...) is 260 MiB of it (NCCL  cuMemImportFromShareableHandle calls)

When the child process loads CUDA it stores some objects on the set GPU, or on the default GPU which is of course GPU 0. So we want to set CUDA_VISIBLE_DEVICES to the current device in each child process before it loads CUDA.

Unfortunately the child process loads CUDA right when it bootstraps the python interpreter, because python immediately imports bergson which uses torch.compile which immediately loads CUDA so it can understand the hardware characteristics for compilation.

So we mutate CUDA_VISIBLE_DEVICES in the parent process before we spawn the copy (this is safe because CUDA_VISIBLE_DEVICES is only read when CUDA is first loaded, so modiying it in the parent has no effect. This is also the reason why we need to absolutely ensure it's set up front in the child processes - we can't fix this in post).

### DIFF
--- a/benchmarks/benchmark_bergson.py
+++ b/benchmarks/benchmark_bergson.py
@@ -29,7 +29,7 @@ from bergson.data import allocate_batches
 from bergson.gradients import GradientProcessor
 from bergson.score.score_writer import InMemorySequenceScoreWriter
 from bergson.score.scorer import Scorer
-from bergson.utils.auto_batch_size import determine_batch_size
+from bergson.utils.batch_size import determine_batch_size
 from bergson.utils.worker_utils import (
     setup_data_pipeline,
     setup_model_and_peft,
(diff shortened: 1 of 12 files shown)


LouisYRYJ:
Yes, I observed the issue with GPU0 needing more memory as well. I didn't manage to fix it, but this seems like it equally distributes all of the load? Amazing 

Edit: Also can we in general squash and merge to avoid clogging of main git history?

luciaquirke:
> Yes, I observed the issue with GPU0 needing more memory as well. I didn't manage to fix it, but this seems like it equally distributes all of the load? Amazing
> 
> Edit: Also can we in general squash and merge to avoid clogging of main git history?

🫡

PR #268, opened 2026-05-03
Decouple MAGIC batch padding quirks from other algorithms

luciaquirke:
bergson's bin-packer enforces a per-rank token budget max_len * batch_size <= N. 

[PR #227](https://github.com/EleutherAI/bergson/pull/227) added an all_reduce(MAX) on max_len in pad_and_tensor to make activation shapes identical across ranks for the MAGIC trainer. But build index also calls pad_and_tensor. 

If there is a large variance in sequence length so there are some 2-token items and some long ones you can get one device that's like [1, 2048] and another that's like [1024, 2] and then when you pad everything to the max length you get a batch that's like [1024, 2048] and everything hangs (crashes on one rank). I'm making this flag-controlled so we can keep your MAGIC fix and turn it off for everything else. But this isn't a great long term solution (due to my limited understanding of the relevant factors).

Adds a CPU test that fails on the pre-fix code path (rank 1 padded to 79,000 tokens vs. 2048 budget).

I don't love this solution - MAGIC will also OOM in scenarios with high sequence length variance. But apparently without it we get hangs? hoping for some more info from @norabelrose

### DIFF
--- a/bergson/collector/collector.py
+++ b/bergson/collector/collector.py
@@ -707,10 +707,14 @@ class CollectorComputer:
 
                 # Compute padded tensors and valid_mask before entering context
                 # TODO check if valid_mask has bug
+                # Local padding only: bin-packer enforces a per-rank
+                # ``max_len × batch_size ≤ N`` budget that a global-max
+                # all-reduce would silently violate.
                 x, y, valid_mask = pad_and_tensor(
                     batch["input_ids"],
                     labels=batch.get("labels"),
                     device=self.device,
+                    sync_max_len=False,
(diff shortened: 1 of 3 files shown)


luciaquirke:
@norabelrose Could you please review this? I think some changes to the batching happened that I wasn't across, and I'd like to confirm that we haven't introduced any bugs. Maybe we need more test coverage now there are multiple variables:

- attribution method
- bin batching vs. chunking
- FSDP vs. DDP (I don't _think_ this affects anything?)

PR #273, opened 2026-05-05
kfac support by setting ev_correction True/False in the HessianCfg

LouisYRYJ:
This PR introduces full support for the EKFAC pipeline (which includes shampoo and tkfac). In parts introduced here already https://github.com/EleutherAI/bergson/pull/271

PR #275, opened 2026-05-05
Add first-class K-FAC support to build

jammastergirish:
[Linear task](https://linear.app/eleutherai-interpretability/issue/ELE-24/add-first-class-k-fac-support-to-build)

## Summary

Compress the K-FAC IVHP output to match `bergson build`'s compressed gradient store. K-FAC factors are computed on uncompressed gradients (unchanged from `main` + #273); the compression happens at apply time as a post-projection of the IVHP output:

```
G̃_q = P_S · (H⁻¹ G_q) · P_Aᵀ      ∈ [N, p, p]
```

`P_S` and `P_A` are the same per-layer Rademacher / normal matrices (`f"{name}/left"`, `f"{name}/right"`) that `bergson build` uses on training gradients (via the existing `create_projection_matrix(...)` helper). Train-side stored gradients are `P_S · G_t · P_Aᵀ` by construction, so their Frobenius inner product with the compressed query approximates `<G_q, H⁻¹ G_t>` per layer up to a known `(p/d)²` constant.

`IndexConfig.projection_dim` (the existing knob that controls compression in build/score) now also drives apply-time K-FAC compression; no new user-facing flag.

## Diff

Net **+16 lines** vs `kfac-full-support` (#273):

- `apply_hessian.py` — two `EkfacConfig` fields, a `grad_sizes` ternary, and a 7-line post-projection block at the end of `compute_ivhp_sharded`. The legacy rotate-divide-rotate above it is byte-identical to `main`.
- `pipeline.py` — thread `index_cfg.projection_dim` / `projection_type` into `EkfacConfig` (step 3), and remove the `score_index_cfg.projection_dim = 0` override (step 4) so the training-side gradient store at score time is built at `[p, p]` to match the apply output.

`HessianConfig` is unchanged — no new fields. `bergson hessian` is unchanged.

## Composability

- Works with `ev_correction=True` — the post-projection step is independent of which lambda the rotate-divide-rotate uses.
- Works for any Kronecker-factored method (`kfac`, `tkfac`, `shampoo`) — the apply path consumes `Q` and `λ` that `compute_eigendecomposition` already writes for all of them.
- One-sided application (per the brief): `H⁻¹` only ever touches the query; training gradients flow straight from build through to score with no inverse-Hessian application.

## Known caveat (pre-existing)

K-FAC + `include_bias=True` has a pre-existing shape mismatch — filed as #277. Out of scope here; this PR doesn't make it worse.

## Test plan

Validated end-to-end on Pythia-14m / pile-10k (200 train docs, 10 query docs, 24 modules):

- [x] `bergson hessian --method kfac` baseline — unchanged from `main` + #273.
- [x] `bergson ekfac --projection_dim 16 --projection_target per_module` — `kfac_query/info.json` shows per-module `grad_sizes = 256 = 16²`, `scores/processor_config.yaml` carries the propagated projection settings, scores produced cleanly.
- [x] `bergson ekfac --projection_dim 0` — legacy regression: full `d_S · d_A` per-module sizes, scores produced.
- [x] `bergson ekfac --projection_dim 16 --ev_correction true` — EV correction + compression compose (no guard needed).


bergson/config.py
```diff
@@ -548,6 +548,17 @@ class HessianConfig(Serializable):
     """Whether to use dataset labels for Hessian (empirical Fisher) approximation.
     If false, the model predictions will be used."""
 
+    projection_dim: int = 0
+    """If > 0, compress the K-FAC factors (A and S) and the EK-FAC eigenvalue
```
luciaquirke: You can check with Louis but I believe we don't support EK-FAC compression, because we can't do this operation with the EK-FAC eigenvalue. That's why we are only supporting K-FAC. So we would want to remove this additional logic.

bergson/hessians/hessian_approximations.py
```diff
@@ -212,7 +232,11 @@ def collect_hessians(
     }
     desc = f"Approximating Hessians with {hessian_cfg.method}"
     if ev_correction:
-        collector = LambdaCollector(**collector_args)
+        collector = LambdaCollector(
+            **collector_args,
+            projection_dim=hessian_cfg.projection_dim,
+            projection_type=hessian_cfg.projection_type,
+        )
```
luciaquirke, line 220: I think this might be backward - surely we only do it when there's no EV correction?

luciaquirke:
@LouisYRYJ you don't need to read the PR description but if you could double check my implementation review that would be fantastic

LouisYRYJ:
Thanks, will check now.
Can we either merge https://github.com/EleutherAI/bergson/pull/273 first or make this PR on top that one?

bergson/config.py
LouisYRYJ: Yes this is correct, we can only do K-FAC in an efficient way

bergson/hessians/hessian_approximations.py
```diff
@@ -145,6 +145,20 @@ def hessian_worker(
 
     model, target_modules = setup_model_and_peft(index_cfg)
 
+    if hessian_cfg.projection_dim > 0:
```
LouisYRYJ: This is not necessary, shampoo or tkfac (though tkfac implementation is most likely wrong and will be removed soon) would also work, as would any Kronecker factored option

LouisYRYJ:
Ok I think this is almost correct, but importantly the compression should be the *last* step in our kfac pipeline. Here is the sketch:
1. Compute KFAC as usual including its eigenvalues https://github.com/EleutherAI/bergson/pull/273 Here we should add an option where save the eigenvalues of A and S (the covariance matrices), so that on disk we have Q_A, Q_S, the eigenvectors, and E_A, E_S, the respective eigenvalues. We have the equality A= Q_A E_A Q_A.T and S= Q_S E_S Q_S.T (or A= Q_A.T E_A Q_A resp. S not sure right now, but in this case it doesnt matter because A and S are symmetric). 
2. Compute the (damped) inverses (E_A+lambda)^{-1} and compute the matrix A'=(A+lambda *id)^-1= Q_A (E_A+lambda)^{-1} Q_A.T. Likewise for S.
3. Now this**this** is the matrix we random project: A'_P= A'*P where P is the Rademacher or whatever random projection. shape[A'_P]=shape[P] and so this will be our new in-place curvature + random projection operation all in one go. Same story for S to get the other side of the projection

This works for any method, once we are provided with A and S, regardless of how they were computed. Eigenvalue correction will not work because it does not only operate on either rows or columns (whereas A, S and their decompositions do).

jammastergirish:
While digging through the apply path during the redesign, Claude noticed an issue. I'm working through everything now and am not fully certain of it, so will discuss/let you know shortly:

K-FAC + `include_bias=True` is broken on `main` (and therefore in this PR too). The K-FAC activation covariance is built without the bias column, so the per-layer reshape in `compute_ivhp_sharded` fails for any layer where bias is being collected.

Filed separately as #277 so the fix doesn't get tangled with the compression work here. This PR's compression path inherits whatever fix lands there for free, since it consumes covariances by their declared dimension.

LouisYRYJ:
Yes, I never implemented support for bias. IMO will not be load bearing, but could be nice for sake of completeness (and technically shampoo or others also dont support it)

bergson/hessians/hessian_approximations.py
```diff
@@ -194,6 +211,39 @@ def hessian_worker(
         world_size=world_size,
     )
 
+    if hessian_cfg.projection_dim > 0:
```
luciaquirke: I don't think you need to change the code for compute Hessians at all. in the apply_hessian pipeline, we randomly project the pre-computed hessian. Then we randomly project the query. Then we do A @ query @ B. the index config with the projection information is likely already present in that pipeline.

bergson/hessians/eigenvectors.py
```diff
@@ -420,3 +430,128 @@ def _gather_and_shard_along_dim0(
         gc.collect()
 
     return result_dict
+
+
+def compute_whitening_projection_matrices(
```
luciaquirke: I am a bit confused but I suspect that basically none of this code is necessary. I think this should be like a 10 line change or something.

bergson/hessians/apply_hessian.py
```diff
@@ -71,8 +75,10 @@ def compute_ivhp_sharded(self):
             eigen_g[k] = eigen_g[k].to(dtype=torch.float32)
             lambda_factor[k] = v.to(dtype=torch.float32)
 
+        p = self.cfg.projection_dim
         grad_sizes = {
-            name: eigen_g[name].shape[1] * eigen_a[name].shape[1] for name in eigen_a
+            name: p * p if p > 0 else eigen_g[name].shape[1] * eigen_a[name].shape[1]
+            for name in eigen_a
```
luciaquirke, line 82: Nice

bergson/hessians/apply_hessian.py
```diff
@@ -143,6 +149,18 @@ def compute_ivhp_sharded(self):
         del eigen_a
         gc.collect()
 
+        if p > 0:
+            pt = self.cfg.projection_type
```
luciaquirke: ideally this would be projection_type, too many initialisms makes for unreadable code. This part of the code base especially suffers from initialisms

jammastergirish:
Agree. Will fix. I’m always saying that to coders so glad you picked me up
on it!

luciaquirke (approved):
🔥 🔥

bergson/hessians/apply_hessian.py
```diff
@@ -29,6 +31,8 @@ class EkfacConfig:
     `HessianConfig.ev_correction=True`."""
     debug: bool = False
     lambda_damp_factor: float = 0.1
+    projection_dim: int = 0
+    projection_type: Literal["normal", "rademacher"] = "rademacher"
 
 
 class EkfacApplicator:
```
LouisYRYJ, line 38: TODO: This should allow us to also do H-1/2 multiplications, this is needed for KFAC + random projection.

bergson/hessians/pipeline.py
LouisYRYJ, line 43: TODO: throw error when we do compression + ev correction. We right now only support kfac without eigenvalue correction.

LouisYRYJ: Also this will work not only for kfac, but other methods too (e.g. shampoo)

LouisYRYJ, line 115: Step 3.5: Compute R @ H^-1/2.

LouisYRYJ: THis needs to be computed for the right and the left projection seperately and you do this by multiplying left_proj with (act or grad cov inverse) and then right_proj with (grad or act cov inverse)

LouisYRYJ: Save to disk

LouisYRYJ, line 126: go in here, and go deep into the logic to find out where random projections are multiplied with the grads. And then add a config or whatever that realizes that we should use the modified random projections we just computed and saved to disk

bergson/hessians/apply_hessian.py
luciaquirke: (Follow up PR)

PR #276, opened 2026-05-05
Unify K-FAC and autocorrelation Hessian save layout

jammastergirish:
[Linear task](https://linear.app/eleutherai-interpretability/issue/ELE-34/ensure-consistent-hessian-approximation-serialization)

## Summary

`bergson hessian` had two different save layouts depending on the method:

- K-FAC family (`kfac` / `tkfac` / `shampoo`) wrote to `run_path/<method>/`.
- Autocorrelation wrote to `run_path/` (because it piggy-backs on `bergson build`).

Both already serialize a `hessian_config.yaml` recording the method, but at different relative locations. This PR aligns them under the K-FAC convention so the saved `hessian_config.yaml` always lives at `<run_path>/<method>/hessian_config.yaml` and unambiguously identifies which method produced the directory.

### Layout

| Path | Before | After |
|---|---|---|
| `bergson hessian --method kfac runs/foo` | `runs/foo/kfac/` | `runs/foo/kfac/` (unchanged) |
| `bergson hessian --method tkfac runs/foo` | `runs/foo/tkfac/` | `runs/foo/tkfac/` (unchanged) |
| `bergson hessian --method shampoo runs/foo` | `runs/foo/shampoo/` | `runs/foo/shampoo/` (unchanged) |
| `bergson hessian --method autocorrelation runs/foo` | `runs/foo/` | **`runs/foo/autocorrelation/`** |

### Where the routing lives

K-FAC-family writes mutate \`run_path\` inside \`approximate_hessians\` (\`bergson/hessians/hessian_approximations.py\`). The autocorrelation flow piggybacks on \`build()\`, which now takes an optional \`hessian_cfg\` parameter; when supplied alongside \`skip_index=True\`, \`build\` performs the same \`<run_path>/<method>/\` mutation and serializes the supplied \`HessianConfig\` to \`hessian_config.yaml\`. The \`Hessian\` CLI command no longer rewrites \`index_cfg.run_path\` itself — it just passes \`hessian_cfg\` through to \`build\`. \`trackstar.py\` and \`bergson/hessians/pipeline.py\` continue to call \`build()\` without a \`hessian_cfg\`, so their custom subdir layouts are unchanged.

### Reading back the method

The canonical way to identify the method that produced a saved directory is to load the on-disk config directly:

\`\`\`python
from bergson.config import HessianConfig

method = HessianConfig.load_yaml(str(path / \"hessian_config.yaml\")).method
\`\`\`

\`mix_autocorrelation_matrices\` uses this same one-liner inline (loop over \`query_path\` and \`index_path\`) — no extra helper is kept, since it was only needed in two places.

## What is not changed

- **On-disk artifact formats are kept as-is.** K-FAC family writes sharded safetensors of Kronecker factors (+ EK-FAC eigenvalue corrections); autocorrelation writes \`GradientProcessor.save\` artifacts (\`hessians.pth\`, \`hessians_eigen.pth\`, \`processor_config.yaml\`, \`normalizers.pth\`). They encode different math and converging the formats would be a large refactor with no clear win.
- **\`bergson build\` (with \`skip_hessians=False\`) still writes its autocorrelation Hessian alongside the gradient index at \`run_path/\`.** That mode produces an index *and* an inline Hessian; namespacing the Hessian under \`<method>/\` would split the run.
- **\`trackstar.py\` and \`bergson/hessians/pipeline.py\`** call \`build()\` / \`approximate_hessians()\` directly with their own custom subdirectories (\`value_hessian/\`, \`query_hessian/\`, etc.); those layouts are unchanged.

## Breaking change

Users invoking \`bergson hessian --method autocorrelation\` and then pointing a downstream \`--preprocess_cfg.hessian_path\` at the run directory must update the path to include the trailing \`/autocorrelation\` segment, e.g.:

\`\`\`diff
  bergson hessian runs/foo --method autocorrelation
- bergson score ... --hessian_path runs/foo
+ bergson score ... --hessian_path runs/foo/autocorrelation
\`\`\`

This matches the K-FAC convention which has always required \`--hessian_method_path runs/foo/kfac\`.

## Verification

Ran on the cluster against Pythia-14m / pile-10k:

- [x] \`bergson hessian runs/auto_unified --method autocorrelation\` lands artifacts at \`runs/auto_unified/autocorrelation/\` with \`hessian_config.yaml\` reporting \`method: autocorrelation\`. No stray files at the run-path root.
- [x] \`bergson hessian runs/kfac_unified --method kfac --ev_correction true\` still produces the same \`runs/kfac_unified/kfac/\` layout (sharded safetensors + configs); no regression.
- [x] \`HessianConfig.load_yaml(str(path / \"hessian_config.yaml\")).method\` returns the right method (\`autocorrelation\` / \`kfac\`) for a saved directory.
- [x] \`mix_autocorrelation_matrices\` continues to validate that both sides are autocorrelation hessians via the inlined check.


bergson/hessians/io.py
```diff
@@ -0,0 +1,42 @@
...
+for autocorrelation Hessian loading.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from bergson.config import HessianConfig
+
+HESSIAN_CONFIG_FILENAME = "hessian_config.yaml"
```
luciaquirke: Let's just use this by value like how we handle the other yaml file names

bergson/hessians/io.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def load_hessian_config(path: str | Path) -> HessianConfig:
```
luciaquirke: Let's just use `HessianConfig.load_yaml(str(cfg_path))` alone so we can avoid having this new file. Same for hessian_method, we don't need it

bergson/__main__.py
```diff
@@ -75,7 +75,23 @@ def execute(self):
 
 @dataclass
 class Hessian(Serializable):
-    """Approximate Hessian matrices using KFAC or EKFAC."""
+    """Approximate Hessian matrices using KFAC or EKFAC.
+
+    Output layout (consistent across methods)::
```
luciaquirke: ` (consistent across methods)`
is redundant

bergson/__main__.py
```diff
@@ -75,7 +75,23 @@ def execute(self):
...
+
+        <run_path>/<method>/
+            hessian_config.yaml   # records ``method`` and other settings
```
luciaquirke: Don't mention method specifically

luciaquirke:
Seems like there's some claude bloat in this implementation, could you please do a pass for conciseness before getting human reviews? I want to avoid the code base getting too big and complex

bergson/hessians/io.py
jammastergirish: Yes, totally right. I should have caught this as I hate non-DRY code.

bergson/__main__.py
```diff
@@ -75,7 +75,15 @@ def execute(self):
 
 @dataclass
 class Hessian(Serializable):
-    """Approximate Hessian matrices using KFAC or EKFAC."""
+    """Approximate Hessian matrices using KFAC or EKFAC.
+
+    Output layout:
```
luciaquirke: This is an implementation deatil and shouldn't live in the docstring

bergson/__main__.py
```diff
@@ -86,6 +94,9 @@ def execute(self):
         validate_run_path(self.index_cfg)
 
         if self.hessian_cfg.method == "autocorrelation":
+            self.index_cfg.run_path = (
+                self.index_cfg.run_path + f"/{self.hessian_cfg.method}"
+            )
```
luciaquirke: Can we instead update the site where the Hessian yaml is written in build to point to the correct location? This seems like potentially a hack

bergson/process_grads.py
```diff
@@ -12,14 +12,10 @@
...
-
-    hessian_cfg = HessianConfig.load_yaml(cfg_path)
-    assert hessian_cfg.method == "autocorrelation", (
-        f"Hessian at '{path}' was computed with method "
-        f"'{hessian_cfg.method}'; mix_autocorrelation_matrices only "
-        f"supports autocorrelation."
+    method = HessianConfig.load_yaml(str(path / "hessian_config.yaml")).method
+    assert method == "autocorrelation", (
+        f"Hessian at '{path}' was computed with method '{method}'; "
+        f"mix_autocorrelation_matrices only supports autocorrelation."
```
luciaquirke: This function can probably be inlined now unless it's used in 3+ places

bergson/build.py
```diff
@@ -137,15 +138,21 @@ def build(
...
 
+    if hessian_cfg is not None and index_cfg.skip_index:
+        index_cfg.run_path = index_cfg.run_path + f"/{hessian_cfg.method}"
+
     index_cfg.partial_run_path.mkdir(parents=True, exist_ok=True)
     index_cfg.save_yaml(index_cfg.partial_run_path / "index_config.yaml")
     preprocess_cfg.save_yaml(index_cfg.partial_run_path / "preprocess_config.yaml")
     if not index_cfg.skip_hessians:
-        HessianConfig(method="autocorrelation").save_yaml(
+        (hessian_cfg or HessianConfig(method="autocorrelation")).save_yaml(
```
luciaquirke, line 155: I think we can just save it if it's not None here? so no `if not index_cfg.skip_hessians:` check

```
if hessian_cfg:
  hessian_cfg.save . . . 
```

bergson/process_grads.py
```diff
@@ -95,8 +82,12 @@ def mix_autocorrelation_matrices(
...
     output_path = Path(output_path)
 
-    assert_autocorrelation_hessian(query_path)
-    assert_autocorrelation_hessian(index_path)
+    for path in (query_path, index_path):
+        method = HessianConfig.load_yaml(str(path / "hessian_config.yaml")).method
+        assert method == "autocorrelation", (
+            f"Hessian at '{path}' was computed with method '{method}'; "
+            f"mix_autocorrelation_matrices only supports autocorrelation."
+        )
```
luciaquirke, line 90: Nice

luciaquirke (approved):
Awesome, LGTM with YAML saving logic fix

bergson/build.py
```diff
@@ -137,15 +138,21 @@ def build(
...
         and aggregation.
+    hessian_cfg : HessianConfig | None
+        When provided and ``index_cfg.skip_index`` is True, the run path is
+        nested under ``<run_path>/<method>/``.
     """
     if index_cfg.debug:
         setup_reproducibility()
 
+    if hessian_cfg is not None and index_cfg.skip_index:
+        index_cfg.run_path = index_cfg.run_path + f"/{hessian_cfg.method}"
```
luciaquirke, line 149: Let's update this to not mutate the run path, in line with how autocorrelation Hessians are saved.

luciaquirke (commented):
I think I'm going to pause this one on reflection, I think the unification needs more thought because it's not really clear which run path strategy we want to commit to between k-fac and autocorrelation

luciaquirke:
cc @LouisYRYJ 

I spoke to Louis and I think he's happy to use the run path strategy we use for autocorrelation. So we should update the K-FAC family of methods to match. This lets us avoid run path mutation.

I think then this will be ready to merge! And will hopefully make the other PRs conceptually simpler

luciaquirke:
I did this in the big refactor to simplify our config.yaml system, should unblock the other PRs

PR #279, opened 2026-05-08
Usability fixes for evaluation tool

luciaquirke:
New features:
- Add an option to filter empty documents from the validation, since the MAGIC paper uses a dataset with a large fraction of empty documents and we would like this to not bias the validation scores. Filtering empty documents boosts MAGIC and deboosts TrackStar. Disabled by default.
- Add an optional maximum batch size to the bin batching algorithm, because for applications like EK-FAC where the gradients are not compressed this matters for controlling the VRAM usage. Demonstrated its use in the EK-FAC comparison. Disabled by default.
- Add higher_is_better flag to scores config explicitly set to true in our IFs pipelines to avoid confusion around how higher is better for IFs but not MAGIC. This is the strategy LM Eval Harness uses for evaluation results.

Fixes:
- Use higher_is_better flag to fix issue where rho sign would be flipped for IFs being validated (previous comparisons would manually reverse the score in _user_ code!)
- Use layer adapter in module registration code to prevent issue with GPT-2 using convolutions backward as linears.

Nits:
- Assert single scores in validation to prevent confusion (we don't support multi-score validation yet).
- Add comments disambiguating global and local rank everywhere to prevent confusion
- Review chunk_and_tokenize and add comments clarifying the situation: that we always return Doc IDs as a canonical map from rearranged data to original data.
- Specify dim in validation argsort to prevent shape ambiguity.
- Use max heap to maintain sorted invariant in batch allocation - previously split batches would both go back into the list at the end regardless of how large the remainder was. This won't necessarily work for large world size/small final batches, which do pop up in smol validation runs

YAML examples collection update:
- Add replicable comparison YAMLs for several methods using the same MAGIC checkpoint and single item query.
- Update the GPT-2 replication YAML to use a single query since that's the methodology of the MAGIC paper.


bergson/hessians/pipeline.py
```diff
@@ -101,6 +101,7 @@ def _validate(cfg: IndexConfig):
         score_index_cfg.projection_dim = 0
         score_index_cfg.skip_hessians = True
         score_cfg.query_path = transformed_query_path
+        score_cfg.higher_is_better = True
```
luciaquirke, line 104: Moving towards setting fields more explicitly so we don't rely on defaults

examples/magic/gpt2_wikitext.yaml
```diff
@@ -5,20 +5,21 @@ overwrite: true
...
   dataset: Salesforce/wikitext
   subset: wikitext-2-raw-v1
-  split: "train[:4608]"
+  split: "train[100:]"
   chunk_length: 512
 
 query:
   dataset: Salesforce/wikitext
   subset: wikitext-2-raw-v1
-  split: "train[-4608:]"
```
luciaquirke, line 14: Using a single query is what the original MAGIC paper does. This higher batch size gets rho ~= 0.98

PR #285, opened 2026-05-25
Add shuffle seed to MAGIC

SrGonao:
Separates initialization and data order from MAGIC training, by adding a new seed to the config.


norabelrose:
I'm not totally sure this decoupling is needed— did you find that you needed it in your experiments? Also, the doc string for `seed` in this PR seems misleading. It's only getting used to create the subsets for leave-k-out validation and nothing else. If we're going to have two seeds, both of them should have more descriptive names.

luciaquirke:
We were investigating the impact of data order on the results. Reviewing this it seems like we could do this without additional seeds by pipelining the magic and validate commands like [pseudocode]:

```
magic:
  run_path: 0
  seed: 0 
  skip_validate: True
magic:
  run_path: 1
  seed: 1
  skip_validate: True
validate:
  run_path: 0
  seed: 0
validate:
  run_path: 1
  seed: 0
```

@norabelrose is that what you were thinking of?

PR #288, opened 2026-05-25
P-SIFT: precondition-and-sketch via modified projections (Louis's review of #275)

jammastergirish:
Stacked on top of #275 — addresses @LouisYRYJ' comments from last week's discussion.

## What's different from #275

Our discussion and the inline comments asked for the precondition-and-sketch approach instead: compose the random projection with `H^{-1/2}` into a single saved matrix `M = R · cov^{-1/2}`, then use `M` wherever the gradient collector used to sample `R`. Query and training gradients both go through `M`, so `<ĝ_z, ĝ_q> ≈ <g_z, H⁻¹ g_q>` without ever applying `H⁻¹` in low-dim space.

This is as per

image

and 

image

## Comment-to-code map

Pulled from comments in #275 

| Louis's comment | Where it lives |
|---|---|
| Top-level: "save the eigenvalues of A and S" | `eigenvectors.py` now writes `eigval_{activation,gradient}_sharded/` next to the existing eigvec dirs |
| Top-level: "compute (damped) inverses then random-project the curvature itself" | `apply_hessian.py:build_kfac_projections` forms `M = R · Q · diag((E + λ·mean(E))^{-1/2}) · Qᵀ` per (layer, side) |
| Top-level: "works for any method, not just kfac" | Method-agnostic — `build_kfac_projections` reads `eigen_*/eigval_*` which kfac/tkfac/shampoo all produce via the shared `hessian_worker` |
| Top-level: "EV correction will not work" | `pipeline.py` raises immediately on `projection_dim > 0 + ev_correction=True`; secondary guard in `compute_ivhp_sharded` |
| `apply_hessian.py:38` — "TODO: H⁻¹/² multiplications" | `D = (E + damp).clamp_min(...).rsqrt()` → `M = R @ (Q * D) @ Q.T` |
| `pipeline.py:43` — "throw error compression + ev_correction" | Top-of-pipeline guard |
| `pipeline.py:43` — "also works for shampoo" | Method-agnostic; same answer as above |
| `pipeline.py:115` — "Step 3.5: Compute R @ H⁻¹/²" | Labeled `Step 3.5/4` print + `_timed` block + dedicated `build_projections_worker` |
| `pipeline.py:115` — "right and left projection separately" | `SIDE_TO_COV = {"left": "gradient", "right": "activation"}`; build loop produces distinct M_left (with S^{-1/2}) and M_right (with A^{-1/2}) |
| `pipeline.py:115` — "save to disk" | `save_file(..., projection_{side}_sharded/shard_{rank}.safetensors)` |
| `pipeline.py:126` — "find where random projections multiply with grads, add a config to use the saved ones" | New `IndexConfig.kfac_projection_path`; `GradientCollector.setup()` calls `load_kfac_projections` which pre-populates `processor._projection_matrices` so `HookCollectorBase.projection()` returns saved M, not a fresh R |
| Lucia: "too many initialisms (`pt`)" | Renamed throughout; only `"pt"` remaining is the safetensors framework arg |
| Bias support | Not addressed — Louis flagged "not load-bearing" in the earlier thread |

## Verification

Smoke config: gpt2, wikitext `train[:128]` as the K-FAC fitting set, `test[:4]` as the query, `ev_correction=False`.

**Single GPU**
- 13/13 K-FAC tests pass (`pytest tests/ekfac_tests/`).
- End-to-end pipeline runs clean; saves eigenvectors, eigenvalues, M_left, M_right; applies `M·G·Mᵀ`; produces 256 well-formed scores (no NaN, heavy-tailed both signs).
- Spearman ρ against legacy uncompressed (`projection_dim=0`) climbs monotonically:

| p | ρ vs legacy |
|---|---|
| 16 | 0.69 |
| 32 | 0.83 |
| 64 | 0.95 |
| 128 | 0.99 |

Clean convergence to ground truth → math is right; low-p noise is exactly what you'd expect from a 16-dim sketch of 768-dim per-layer matrices.

**Multi-GPU (2× A40)**
Sharded pipeline completes cleanly — exercises `build_kfac_projections`, `fair_distribute_by_cost`, per-rank shard writes, multi-shard load in `_apply_compressed` and `load_kfac_projections`.

| | 1-GPU vs legacy | 2-GPU vs legacy | 1-GPU vs 2-GPU |
|---|---|---|---|
| Legacy (p=0) | — | — | 0.997 |
| p=16 | 0.69 | 0.48 | 0.73 |
| p=32 | 0.83 | 0.89 | 0.91 |
| p=64 | 0.95 | 0.96 | 0.96 |
| p=128 | 0.99 | 0.97 | 0.98 |

- Legacy 1-GPU vs 2-GPU = 0.997 confirms the distributed aggregation pipeline is sound independent of P-SIFT.
- Both P-SIFT runs converge to ground truth as p grows; they also converge to each other.
- Bit-identical scores across GPU counts are *not* guaranteed: FP non-associativity in `dist.all_reduce(SUM)` during covariance accumulation produces a slightly different `M` per layout, and that small difference gets amplified by a low-p sketch. Rank correlation is what matters for retrieval and it's solid by p ≥ 32.

## Note for reviewers

There's one harmless `UserWarning` about a non-writable NumPy array in `apply_hessian.py:259` (the mmap is read-only and we never write to that tensor). Can fix with a `.copy()` if desired; left as-is to keep the diff focused on Louis's comments.

### DIFF
--- a/bergson/collector/gradient_collectors.py
+++ b/bergson/collector/gradient_collectors.py
@@ -72,6 +72,22 @@ class GradientCollector(HookCollectorBase):
         self.lo = torch.finfo(self.save_dtype).min
         self.hi = torch.finfo(self.save_dtype).max
 
+        # When a K-FAC compressed-projection path is provided, pre-populate
+        # the per-layer projection cache with ``M = R · cov^{-1/2}``. The
+        # backward hook's ``self.projection(...)`` call then hits the cache
+        # and uses M in place of a freshly sampled random matrix — so the
+        # gradient comes out preconditioned and sketched in one step.
+        if self.cfg.kfac_projection_path:
+            from bergson.hessians.apply_hessian import load_kfac_projections
+
+            load_kfac_projections(
+                self.cfg.kfac_projection_path,
+                self.processor._projection_matrices,
+                self.target_info.keys(),
+                device=self.model.device,
+                dtype=self.save_dtype,
+            )
+
         self.per_doc_losses = torch.full(
             (len(self.data),),
             device=self.model.device,
--- a/bergson/collector/gradient_collectors.py
+++ b/bergson/collector/gradient_collectors.py
@@ -13,6 +13,7 @@ from torch import Tensor
 from bergson.builder import Builder
 from bergson.collector.collector import HookCollectorBase
 from bergson.config import IndexConfig, PreprocessConfig
+from bergson.hessians.apply_hessian import load_kfac_projections
 from bergson.process_autocorrelation import process_autocorrelation_matrices
 from bergson.score.scorer import Scorer
 from bergson.utils.projection import make_global_projector
@@ -73,8 +74,6 @@ class GradientCollector(HookCollectorBase):
         self.hi = torch.finfo(self.save_dtype).max
 
         if self.cfg.kfac_projection_path:
-            from bergson.hessians.apply_hessian import load_kfac_projections
-
             load_kfac_projections(
                 self.cfg.kfac_projection_path,
                 self.processor._projection_matrices,
--- a/bergson/collector/gradient_collectors.py
+++ b/bergson/collector/gradient_collectors.py
@@ -72,11 +72,6 @@ class GradientCollector(HookCollectorBase):
         self.lo = torch.finfo(self.save_dtype).min
         self.hi = torch.finfo(self.save_dtype).max
 
-        # When a K-FAC compressed-projection path is provided, pre-populate
-        # the per-layer projection cache with ``M = R · cov^{-1/2}``. The
-        # backward hook's ``self.projection(...)`` call then hits the cache
-        # and uses M in place of a freshly sampled random matrix — so the
-        # gradient comes out preconditioned and sketched in one step.
         if self.cfg.kfac_projection_path:
             from bergson.hessians.apply_hessian import load_kfac_projections
 
--- a/bergson/config.py
+++ b/bergson/config.py
@@ -397,6 +397,14 @@ class IndexConfig(AttributionConfig, Serializable):
     processor_path: str = ""
     """Path to a precomputed processor."""
 
+    kfac_projection_path: str = ""
+    """Path to a K-FAC hessian directory containing precomputed modified
+    projection matrices (``projection_left_sharded/``,
+    ``projection_right_sharded/``). When set, the gradient collector loads
+    these instead of sampling fresh random projections — the saved matrices
+    are ``M = R · cov^{-1/2}``, so applying them to gradients yields
+    preconditioned-and-sketched gradients in a single matmul (P-SIFT)."""
+
     optimizer_state: str = ""
     """Source for optimizer second moments used to normalize gradients.
     Either a local path (a checkpoint directory containing ``optimizer.pt``,
--- a/bergson/hessians/apply_hessian.py
+++ b/bergson/hessians/apply_hessian.py
@@ -124,6 +124,12 @@ def build_kfac_projections(
                 device,
                 projection_type,
             )
+            # create_projection_matrix returns unit-norm rows, so E[RᵀR] =
+            # (p/d)·I. An unbiased sketch needs E[RᵀR] = I, so rescale by
+            # √(d/p); otherwise each side carries a p/d factor and the score
+            # picks up a per-layer p²/(d_left·d_right) reweighting that
+            # corrupts ranking relative to the exact (projection_dim=0) path.
+            R = R * (d / projection_dim) ** 0.5
 
             M = R @ (Q * D) @ Q.T
             saved[side][name] = M.to(dtype=dtype).cpu().contiguous()
--- a/bergson/hessians/apply_hessian.py
+++ b/bergson/hessians/apply_hessian.py
@@ -69,9 +69,6 @@ def build_kfac_projections(
     rank = dist.get_rank() if dist.is_initialized() else 0
     world_size = dist.get_world_size() if dist.is_initialized() else 1
 
-    # Discover layer names and per-side dimensions from one shard of each
-    # eigenvector file. The eigendecomp stores Q with shape [d/W, d] per
-    # shard, so the column dim is the full layer dimension.
     side_dims: dict[str, dict[str, int]] = {"left": {}, "right": {}}
     for side, cov in SIDE_TO_COV.items():
         with safe_open(
--- a/bergson/hessians/apply_hessian.py
+++ b/bergson/hessians/apply_hessian.py
@@ -8,13 +8,18 @@ from typing import Literal
 import numpy as np
 import torch
 import torch.distributed as dist
-from safetensors.torch import load_file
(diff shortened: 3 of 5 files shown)


luciaquirke:
> Can fix with a .copy() if desired

Nah not an issue

bergson/collector/gradient_collectors.py
```diff
@@ -72,6 +72,17 @@ def setup(self) -> None:
         self.lo = torch.finfo(self.save_dtype).min
         self.hi = torch.finfo(self.save_dtype).max
 
+        if self.cfg.kfac_projection_path:
+            from bergson.hessians.apply_hessian import load_kfac_projections
```
luciaquirke: imports at the top please

pyproject.toml
```diff
@@ -22,9 +22,17 @@ dependencies = [
     "pyarrow",
     "pyyaml",
     "simple-parsing",
-    "torch>=2.4,<2.10",
+    "torch>=2.5,<2.10",
+    # torchvision/torchaudio aren't used by bergson directly, but transformers'
+    # image_utils does an unconditional `from torchvision.io import ...` at
+    # import time when torchvision is present. RunPod (and most CUDA base
```
luciaquirke, line 28: I don't think we should reference RunPod in particular in our library

.pre-commit-config.yaml
```diff
@@ -8,11 +8,11 @@ repos:
...
         exclude: \.txt$
 -   repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.5.0
+    rev: 26.5.1
     hooks:
     -   id: black
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.15.13'
+    rev: 'v0.15.14'
     hooks:
```
luciaquirke, line 16: Do we need to change this here?

bergson/magic/cli.py
```diff
@@ -93,15 +93,15 @@ def compute_query_gradients(
     if dist.is_initialized():
         if not fsdp:
             grad_accum = {
-                k: differentiable_all_reduce(
-                    g,
-                    "sum",
-                    dist.distributed_c10d._get_default_group(),
+                k: wait_tensor(
```
luciaquirke, line 96: Are these changes intentional? We might need to rebase

LouisYRYJ:
Thanks @jammastergirish!
Looks all correct at first glance, maybe you can rebase before I took a close look?

jammastergirish:
> Thanks @jammastergirish! Looks all correct at first glance, maybe you can rebase before I took a close look?

Oh sorry. I thought I had. Will do shortly.

bergson/hessians/apply_hessian.py
```diff
@@ -35,6 +40,148 @@ class EkfacConfig:
...
+    the eigendecomposition; each rank writes its own shard.
+    """
+    if projection_dim <= 0:
+        raise ValueError(
+            f"build_kfac_projections requires projection_dim > 0 (compression); "
+            f"got {projection_dim}."
+        )
+
+    rank = dist.get_rank() if dist.is_initialized() else 0
+    world_size = dist.get_world_size() if dist.is_initialized() else 1
```
LouisYRYJ, line 76: make sure you use the correct rank and world size (important for multi node setups)

LouisYRYJ: (it seems like they should be the global ones?)

bergson/hessians/apply_hessian.py
```diff
@@ -35,6 +40,148 @@ class EkfacConfig:
...
+                hessian_method_path, f"eigen_{cov}_sharded/shard_0.safetensors"
+            ),
+            framework="pt",
+        ) as f:
+            for name in f.keys():
+                side_dims[side][name] = f.get_tensor(name).shape[-1]
+
+    names = list(side_dims["left"].keys())
+
+    per_layer_dim = {n: max(side_dims["left"][n], side_dims["right"][n]) for n in names}
```
LouisYRYJ, line 91: name is somewhat confusing and I am not sure I understand what this is doing. If you want to fairly distribute by cost, then looking at the max is not necessarily correct?

bergson/hessians/apply_hessian.py
```diff
@@ -35,6 +40,148 @@ class EkfacConfig:
...
+                torch.float32,
+                device,
+                projection_type,
+            )
+            # create_projection_matrix returns unit-norm rows, so E[RᵀR] =
+            # (p/d)·I. An unbiased sketch needs E[RᵀR] = I, so rescale by
+            # √(d/p); otherwise each side carries a p/d factor and the score
+            # picks up a per-layer p²/(d_left·d_right) reweighting that
+            # corrupts ranking relative to the exact (projection_dim=0) path.
+            R = R * (d / projection_dim) ** 0.5
```
LouisYRYJ, line 138: not sure how relevant that is, we look at relative rankings only anyway

LouisYRYJ: I suppose we can leave it there

bergson/hessians/apply_hessian.py
```diff
@@ -53,6 +200,72 @@ def __init__(self, cfg: EkfacConfig, apply_fn=None):
...
+            sketched = torch.einsum("ps,nsa,ra->npr", M_l, G, M_r)
+            grad_buffer[name][:] = (
+                sketched.to(device="cpu", non_blocking=True).flatten(1).numpy()
+            )
+
+        torch.cuda.synchronize() if torch.cuda.is_available() else None
+        grad_buffer.flush()
+        self.logger.info(f"Saved sketched IVHP gradients to {self.cfg.run_path}")
+
+    def _apply_legacy(self):
```
LouisYRYJ, line 267: why is this legacy?

bergson/hessians/pipeline.py
```diff
@@ -93,19 +107,36 @@ def _validate(cfg: IndexConfig):
...
+        if not _step_complete(projections_path, resume):
+            with _timed("step3.5_build_projections", durations):
+                launch_distributed_run(
+                    "build_projections",
+                    build_projections_worker,
+                    [ekfac_cfg],
+                    index_cfg.distributed,
+                )
+    else:
+        # ── Step 3 (legacy): Apply inverse Hessian via rotate-divide-rotate ──
```
LouisYRYJ, line 134: I think all these are not legacy, maybe I am missing something here

bergson/config.py
```diff
@@ -397,6 +397,14 @@ class IndexConfig(AttributionConfig, Serializable):
     processor_path: str = ""
     """Path to a precomputed processor."""
 
+    kfac_projection_path: str = ""
```
LouisYRYJ, line 400: not sure we should even make this configurable? Can't imagine a situation where we would need to change it

LouisYRYJ:
Left a few comments, I think we are almost there!

We should also for now raise Error when doing lora + random projections because in that case one side should not be compressed, but we don't handle that right now. Would be a good follow up PR

PR #290, opened 2026-06-01
Refactor skip hessians, remove processor_path

luciaquirke:
This PR is to simplify the library before release.

Problem 1: We have been using a flag in `IndexConfig` to enable autocorrelation Hessian estimation simultaneously with an index build. We also now have a standalone tool, `bergson hessian`, to enable estimation of many types of Hessians.  We have also begun to use the `IndexConfig` for all config related to gradient collection, including in contexts where Hessians are always estimated (`bergson hessian`) or never estimated (`bergson reduce`). This is convoluted and confusing.

Solution 1: Move this flag into a new `BuildConfig` CLI config class, unavailable to other CLI tools. Set it manually in contexts where the value is constant, e.g. Hessians are always enabled in `bergson hessian` code. This will prevent users from seeing config options that don't make sense.

Problem 2: `processor_path` is confusing and not useful. The original idea was that Hessians and normalizers could be pre-computed and then loaded in and applied at build time, but when the processor path is present it's actually used to compute _new_ Hessians, which is very confusing. We also don't like that this class couples loading normalizers and loading Hessian approximations. We have partially fixed this by adding a standalone `hessian_path` field in PreprocessConfig to apply a precomputed Hessian, and a standalone `optimizer_state` field in IndexConfig to apply a precomputed Adam state buffer. 

Solution 2: Finish the job by removing the processor_path field.

Solution 2 (follow-up PR): Migrate the `optimizer_state` field in IndexConfig to the PreprocessConfig.

### DIFF
--- a/bergson/build.py
+++ b/bergson/build.py
@@ -81,9 +81,12 @@ def build_worker(
         )
 
     model, target_modules = setup_model_and_peft(index_cfg)
-    processor = create_processor(model, index_cfg, hessian_cfg is None, target_modules)
+    skip_hessians = hessian_cfg is None
+    processor = create_processor(model, index_cfg, target_modules)
 
-    maybe_auto_batch_size(index_cfg, model, ds, processor, target_modules, rank)
+    maybe_auto_batch_size(
+        index_cfg, model, ds, processor, target_modules, rank, skip_hessians
+    )
 
     attention_cfgs = {
         module: index_cfg.attention for module in index_cfg.split_attention_modules
@@ -97,6 +100,7 @@ def build_worker(
         "target_modules": target_modules,
         "attention_cfgs": attention_cfgs,
         "preprocess_cfg": preprocess_cfg,
+        "skip_hessians": skip_hessians,
     }
 
     if isinstance(ds, Dataset):
--- a/bergson/build.py
+++ b/bergson/build.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 from datetime import timedelta
+from dataclasses import dataclass
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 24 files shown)


bergson/score/score.py
```diff
@@ -269,7 +269,7 @@ def score_worker(
         )
 
     model, target_modules = setup_model_and_peft(index_cfg)
-    processor = create_processor(model, index_cfg, target_modules)
+    processor = create_processor(model, index_cfg, True, target_modules)
```
LouisYRYJ: also confused by this, what does this argument do?

bergson/build.py
```diff
@@ -74,7 +81,7 @@ def build_worker(
         )
 
     model, target_modules = setup_model_and_peft(index_cfg)
-    processor = create_processor(model, index_cfg, target_modules)
+    processor = create_processor(model, index_cfg, hessian_cfg is None, target_modules)
```
LouisYRYJ: same here, don't see the argument in the original function

LouisYRYJ:
Very cool, happy to see this done. There were two flags about the types that I also don't understand. I think after resolving these, we should be good to go? 

Nit re naming: Not sure if optimizer_state is the appropriate term here?

bergson/score/score.py
luciaquirke: Loads hessians, when it's off it just loads normalizers

bergson/build.py
luciaquirke: it's skip_hessians

bergson/score/score.py
luciaquirke: Outdated, removed

luciaquirke:
> Very cool, happy to see this done. There were two flags about the types that I also don't understand. I think after resolving these, we should be good to go?
> 
> Nit re naming: Not sure if optimizer_state is the appropriate term here?

Not in scope for this PR but let me know if you have a suggestion, I'm planning some more refactors

PR #293, opened 2026-06-03
Consolidate per-run YAMLs into one reproducible config.yaml

luciaquirke:
Each command previously scattered several config files across its run directory (index_config.yaml, preprocess_config.yaml, hessian_config.yaml, score_config.yaml, run_config.yaml, mix_config.yaml). Replace them with a single config.yaml, following the format previously supported using `bergson pipeline`.

Any completed run or sub-command can now be replayed with `bergson <run_dir>/config.yaml`.

- Clean up some Claude slop in the example YAMLs
- Remove run path mutation anti-pattern. This was preventing reproducibility because each time we re-ran the YAML the path would add another `/{method}` suffix.
- Save all the input config to a single config.yaml upfront in each command. This goes to a unique pipeline directory for pipeline runs and the previously existing run directory for each sub-command.
- Refactor the location of the CLI commands to prevent cyclic imports - move out of \_\_main\_\_
- Save single-command config.yamls before calling each command in the CLI logic, rather than in the application logic. This makes it easier to check that all the config is being saved correctly.
- Use YAML to save the derived config processor_config.yaml - this was renamed to use a yaml extension but was still using a json loader. This change is expected to be backwards compatible because JSON is mostly a subset of YAML.


bergson/config/config.py
```diff
@@ -572,7 +572,7 @@ class ApproxUnrollingConfig(Serializable):
 class HessianConfig(Serializable):
     """Config for reducing the gradients."""
 
-    method: Literal["kfac", "tkfac", "shampoo", "autocorrelation"] = "autocorrelation"
+    method: Literal["kfac", "tkfac", "shampoo", "autocorrelation"]
```
luciaquirke, line 575: This can now be included in `build` with `--method "autocorrelation"` to simultaneously collect hessians (previously we had an extra `--skip_hessians` flag, and didn't save any hessian config when we computed the autocorrelation hessian which created inconsistencies). 

If this has a default value it can't be used in this way because of simple_parsing's limitations

bergson/__main__.py
```diff
@@ -284,49 +51,38 @@ def execute(self):
 def main():
     """Parse CLI arguments and dispatch to the selected subcommand.
 
-    Three input shapes are supported:
-      `bergson <command> --flag value ...`  — single-command CLI-flag mode
-      `bergson pipeline <file.yaml>`        — multi-step pipeline mode
-      `bergson <command> <file.yaml|json>`  — single-command config-file mode
+    Two input shapes are supported:
```
luciaquirke, line 80: Simplified this interface from three to two types of inputs, was way too confusing.

bergson/score/score.py
```diff
@@ -68,11 +69,9 @@ def get_query_grads(
...
         grad_sizes = metadata["grad_sizes"]
 
-    preprocess_path = Path(query_path / "preprocess_config.yaml")
-    if preprocess_path.exists():
-        preprocess_cfg = PreprocessConfig.load(preprocess_path)
-    else:
-        preprocess_cfg = PreprocessConfig()
+    preprocess_cfg = load_subconfig(
+        query_path, "preprocess_cfg", PreprocessConfig, default=PreprocessConfig()
+    )
```
luciaquirke, line 75: todo this method feels like claude slop, too many side effect-y things

bergson/query/query_index.py
```diff
@@ -56,7 +57,12 @@ def query(
         Configuration describing the index path, HF model to load, and dataset field
         used to print the retrieved documents.
     """
-    index_cfg = IndexConfig.load_yaml(Path(query_cfg.index) / "index_config.yaml")
+    index_cfg = load_subconfig(query_cfg.index, "index_cfg", IndexConfig)
+    if index_cfg is None:
+        raise FileNotFoundError(
+            f"Could not load an index config from '{query_cfg.index}'. "
+            f"Expected a '{CONFIG_FILENAME}' produced by a build/reduce/hessian run."
+        )
```
luciaquirke, line 65: todo check claudeslop

luciaquirke:
cc @LouisYRYJ

bergson/cli/commands.py
```diff
@@ -0,0 +1,279 @@
...
+from ..magic import MagicConfig, run_magic
+from ..process_grads import mix_autocorrelation_matrices
+from ..query.query_index import query
+from ..score.score import score_dataset
+from ..utils.worker_utils import validate_run_path
+from .config_io import save_run_config
+
+
+@dataclass
+class Build(Serializable):
```
LouisYRYJ, line 64: Sort these alphabetically

bergson/approx_unrolling/precompute_checkpoints.py
```diff
@@ -183,7 +183,7 @@ def precompute_checkpoint_hessians(
         )
 
         ckpt_index_cfg = deepcopy(index_cfg)
-        ckpt_index_cfg.run_path = str(ckpt_dir)
+        ckpt_index_cfg.run_path = str(out_path)
```
LouisYRYJ, line 186: I think this is unrelated to the rewrite? But I think it is correct

bergson/score/score.py
luciaquirke: Going to return T | None and then throw outside the helper if necessary

PR #294, opened 2026-06-05
Fix K-FAC covariance shapes when include_bias=True

### LINKED ISSUE #277: K-FAC Hessian + include_bias=True produces incompatible gradient/covariance shapes
Found/generated by Claude during PR 275.

## Summary

When `bergson build` is run with `include_bias=True` and the Hessian step uses `method='kfac'`, `apply_hessian` (and the scoring paths downstream of it) fails at a per-layer reshape: the gradient store has one extra column per layer that the K-FAC activation covariance doesn't account for.

This is a **pre-existing** limitation independent of the K-FAC compression work in #275 — the same failure happens on `main`. Filing separately so the fix doesn't get tangled with the compression PR.

## What goes wrong (concrete shapes)

For each linear layer `nn.Linear(I, O, bias=True)`:

- **`bergson build`** with `include_bias=True` stores per-sample gradients of shape `[O, I+1]` per layer — the bias gradient is concatenated as an extra "activation" column. (`HookCollectorBase.shapes()` at `bergson/collector/collector.py:264-270` sets `grad_shape[-1] += 1` when `collect_bias`; `_compute_gradient` does the matching `torch.cat`.)
- **K-FAC `CovarianceCollector`** at `bergson/hessians/kfac.py` computes `A = aᵀa` from the raw forward input `a: [N·S, I]` — **no bias column**. The `collect_bias` flag is unpacked in `_init_covariance_dict` (`bergson/hessians/sharded_computation.py:28`) but never used. Result: `A: [I, I]`.
- At apply time, `compute_ivhp_sharded` reshapes the loaded query gradient via:
  ```python
  gradients_noi.view(-1, eigen_g[k].shape[1], eigen_a[k].shape[1])
  # = view(-1, O, I)
  # but stored flat size is N·O·(I+1) → reshape error
  ```

## Repro

```sh
bergson build --processor.include_bias true ...
bergson approximate-hessians --hessian_cfg.method kfac ...
bergson apply-hessian ...    # raises at .view(-1, O, I)
```

## Fix sketch

Teach K-FAC's covariance collection to operate on the **augmented activation** `[a; 1]` of shape `[N·S, I+1]` when the layer's bias is being collected:

- `CovarianceCollector.forward_hook` appends a 1-column to `a` (matching the build-time gradient layout) before `aᵀa`, giving `A: [I+1, I+1]`.
- `_init_covariance_dict` sizes the activation covariance as `[I+1, I+1]` when `collect_bias=True`.
- Downstream (`compute_eigendecomposition`, `compute_whitening_projection_matrices`, `apply_hessian`) all derive `d_A` from `A.shape[-1]`, so they pick up the new dimension automatically.

The gradient covariance `S: [O, O]` is unchanged.

## Scope

- Out of scope for #275 — that PR is about compression and inherits whatever fix lands here for free.
- In scope for whoever picks this up: a fix here also unblocks K-FAC + bias on `main` (legacy IVHP path).

luciaquirke:
Closes #277.

`bergson build` with `include_bias=True` stores per-layer gradients of shape `[O, I+1]` (the bias gradient is appended as an extra "activation" column), but the K-FAC covariance path computed `A = aᵀa` on the raw activation, giving `A: [I, I]`. `apply_hessian`'s per-layer `view(-1, O, I)` then failed on the flat `N·O·(I+1)` gradients:

```
RuntimeError: shape '[-1, 128, 128]' is invalid for input of size 16512
```

reproduce with `bergson ekfac --include_bias true --method kfac --model EleutherAI/pythia-14m`.

## Fix

- `collect_hessians` now passes `GradientProcessor(include_bias=index_cfg.include_bias)` to the Hessian collectors — previously the `collect_bias` flags in `target_info` were always `False` during Hessian fitting regardless of the build config.
- `_init_covariance_dict` sizes the activation covariance `[I+1, I+1]` when `collect_bias`.
- `CovarianceCollector` (kfac), `TraceCovarianceCollector` (tkfac), and `ShampooCollector` augment the activation with a ones column, matching the build-time gradient layout.
- `LambdaCollector` augments the activation before rotating by `eigen_a` (now `[I+1, I+1]`), so EKFAC eigenvalue corrections come out `[O, I+1]`. 
- Sharded computation now supports unevenly divisible dimensions via `shard_bounds` (rank 0 absorbs the remainder) (in this PR because with the bias column, `I+1` is almost never divisible by the world size.)
- Fixed `step2_fit_hessian` timing bug (`approximate_hessians` ran outside its `_timed` block, always reporting 0.0s).

### DIFF
--- a/bergson/hessians/sharded_computation.py
+++ b/bergson/hessians/sharded_computation.py
@@ -6,6 +6,21 @@ from torch import Tensor
 from bergson.utils.utils import get_device
 
 
+def shard_bounds(dim: int, rank: int, world_size: int) -> tuple[int, int]:
+    """Range [start, end) of ``rank``'s shard of a dimension of size
+    ``dim`` split across ``world_size`` ranks.
+
+    Rank 0 takes the remainder rows when ``dim`` is not evenly divisible,
+    so the shard sizes are [base + remainder, base, ..., base].
+    """
+    base, remainder = divmod(dim, world_size)
+    if rank == 0:
+        return 0, base + remainder
+
+    start = remainder + rank * base
+    return start, start + base
+
+
 class ShardedMul:
     def __init__(
         self,
@@ -15,6 +30,10 @@ class ShardedMul:
         self.rank = dist.get_rank() if self.dist else 0
         self.world_size = dist.get_world_size() if self.dist else 1
         self.device = torch.device(get_device(self.rank))
+
+    def shard_bounds(self, dim: int, rank: int | None = None) -> tuple[int, int]:
+        """Row range [start, end) of ``rank``'s shard (default: this rank)."""
+        return shard_bounds(dim, self.rank if rank is None else rank, self.world_size)
 
     def _init_covariance_dict(
         self,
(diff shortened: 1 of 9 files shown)


bergson/hessians/sharded_computation.py
```diff
@@ -6,6 +6,21 @@
 from bergson.utils.utils import get_device
 
 
+def shard_bounds(dim: int, rank: int, world_size: int) -> tuple[int, int]:
```
LouisYRYJ, line 9: any reason we have this basically twice?

LouisYRYJ:
I am really happy to see the uneven shard/dim issue fixed. This looks great and we can merge it.

bergson/hessians/sharded_computation.py
luciaquirke: If you mean the inner shard function in the class Claude suggested this pattern, I guess to make it more clear which variables are constant within the class and which are defined in the function call flow.
I'm currently not sure how much to "desloppify" claude code to bring it in line with my coding style, like I imagine at least some of its decisions are because that's how the Anthropic engineers prefer it and maybe they know something I don't? e.g. I'm coming around to using underscore prefixes for private methods in line with standard python recommendations, even though it's ugly, because it does seem like it makes the code a bit more clear. I do think its code is incredibly bloated and has too many layers of indirection so also happy to get rid of this one in a follow up if you have a strong take.

PR #300, opened 2026-06-16
Grad similarity test for completion dataset: Token scores (sum) vs Sequence score + Double BOS bug fix

sweta-98:
- Tests if Token scores (sum) == Sequence score for 1 sample.
Run bergson_scrs_tok_vs_seq_validate_1sample.sh

- Double BOS fix in tokenize function + print statements for debugging


sweta-98:
Currently fails-
---------------
Sequence score: -9.000981
Token scores sum: -0.040792614
Token scores mean: -0.0027195076
---------------
---> TEST FAILED

luciaquirke:
Okay so per-token currently only saves gradients over the completion and not the prompt. For it to sum you would also want the prompt scores here - these positions will just have grads that are flowing back from the later loss-bearing tokens. Going to include those by default. Note that it's kind of unclear how you would re-weight these tokens because they already won't have losses. Perhaps we want to explore adding a gradient mask that can re-weight flow-back grads? Interested in your thoughts on this, seems like novel territory.

luciaquirke:
Does this work for you? https://github.com/EleutherAI/bergson/pull/322

luciaquirke:
This should be fixed now, but you might need to use

```
 np.allclose(ex_scores.sum(), seq_scrs[i], rtol=1e-3, atol=1e-3)
```

to avoid getting a false error from floating point errors

PR #301, opened 2026-06-22
Remove skip_index and simultaneous hessian/build

luciaquirke:
We've had this efficiency optimization where the autocorrelation hessian can be computed simultaneously with the index build, but it's a bit convoluted and my understanding is that the amount of data you need for a good autocorrelation estimate is quite low anyway, so the speed gains are limited. Removing for simplicity.

### DIFF
--- a/bergson/hessians/hessian_approximations.py
+++ b/bergson/hessians/hessian_approximations.py
@@ -14,6 +14,7 @@ from bergson.config.config import AttentionConfig, HessianConfig, IndexConfig
 from bergson.data import allocate_batches
 from bergson.distributed import init_dist, launch_distributed_run
 from bergson.gradients import GradientProcessor
+from bergson.hessians.autocorrelation import AutocorrelationCollector
 from bergson.hessians.eigenvectors import (
     LambdaCollector,
     compute_eigendecomposition,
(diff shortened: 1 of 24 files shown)


bergson/hessians/hessian_approximations.py
```diff
@@ -128,16 +130,42 @@ def hessian_worker(
     if target_modules is None:
         target_modules = peft_target_modules
 
+    attention_cfgs = {
+        module: index_cfg.attention for module in index_cfg.split_attention_modules
+    }
+    batches = allocate_batches(ds["length"][:], index_cfg.token_batch_size)
+
+    # The autocorrelation Hessian is a dense per-module gradient Gram so
+    # it computes in one pass and skips the factored eigendecomposition
```
LouisYRYJ, line 139: Hmmh I think it may be better to modify approxmiate hessians to control whether we do EV computations

LouisYRYJ:
nice, this looks good! I think we can probably unify the hessian approaches even more, but that may be done in another PR. Also important to be aware that everything outside of the autocorr hessian uses Fisher and not empirical Fisher (i.e. uses sampled labels from the logits, not actual data labels)

luciaquirke:
Yeah I'd appreciate if you could handle any final Hessian code simplifications, I'm probably not as across this part of the codebase

luciaquirke:
I'm going to make one more PR trying to improve the hessian inversion stuff a bit

luciaquirke:
I might still need help getting it to full unification but I have a follow up here: https://github.com/EleutherAI/bergson/pull/304

PR #303, opened 2026-06-29
Add global-norm gradient clipping (max_grad_norm) to MAGIC trainer

davidoj:
## Add global-norm gradient clipping (`max_grad_norm`) to the MAGIC trainer

**Written by Claude.**

### What

Adds an optional `max_grad_norm: float | None = None` to `TrainingConfig`. When
set, gradients are clipped to that global norm before each optimizer step, matching
HuggingFace `Trainer`'s `max_grad_norm`. `None`/`0` disables clipping and preserves
the current behavior exactly.

This lets a MAGIC replay faithfully match training runs that clip — previously the
MAGIC trainer had no clipping and no way to enable it.

### How

One config field, one clip step in `Trainer.step()` (applied between the cross-rank
gradient reduce and `self.optimizer.update`), and the value threaded from config
into the trainer constructor.

```python
sq_norm = sum(g.pow(2).sum() for g in grads.values())
# ... reduce across FSDP shards if needed ...
coef = (self.max_grad_norm / (sq_norm.sqrt() + 1e-6)).clamp(max=1.0)
grads = {k: g * coef for k, g in grads.items()}   # out-of-place under trace
```

### MAGIC-specific subtleties

MAGIC differentiates the training trajectory w.r.t. implicit per-token data weights,
so the clip must stay inside the autograd graph or influence scores silently become
approximate (the same failure class as the historical non-differentiable all-reduce
bug). Two points:

- **Out-of-place under `trace=True`.** The clip uses `g * coef`, not `g.mul_(coef)`,
  because an in-place update breaks the graph MAGIC backprops through. The non-traced
  forward replay still clips in-place to save memory.
- **Global norm across all params *and* all ranks/FSDP shards.** Under FSDP the grads
  are sharded `DTensor`s, so the squared norm is a partial sum over each rank's shard;
  it's redistributed to `Replicate` (an autograd-aware all-reduce) rather than a plain
  `dist.all_reduce`, which raises on `DTensor`s. In the DDP path the grads are already
  replicated by the all-reduce above, so the local sum is already global.

### Tests

- `tests/test_grad_clipping.py`
  - a biting `max_grad_norm` scales an SGD update by exactly the clip coef;
  - `None` and a non-biting threshold reproduce the unclipped trajectory bit-for-bit;
  - clipped MAGIC scores are finite, nonzero, and change the trajectory;
  - **MAGIC influence through the clip matches a finite-difference gradient** of the
    query loss w.r.t. the per-doc weights — the check that fails if the clip ever
    leaves the autograd graph.
- `tests/test_distributed_magic.py::test_fsdp_ddp_scores_match_with_grad_clipping`
  (2 GPUs): FSDP and DDP produce matching clipped scores, exercising the cross-shard
  norm reduction. The unclipped run is the control that the threshold actually bites;
  checks are scale-free since scores are ~1e-6.

### DIFF
--- a/bergson/config/config.py
+++ b/bergson/config/config.py
@@ -346,6 +346,10 @@ class TrainingConfig(AttributionConfig, Serializable):
     weight_decay: float = 0.01
     """Weight decay coefficient for AdamW and Muon."""
 
+    max_grad_norm: float | None = None
+    """Clip gradients to this global norm before each optimizer step, matching
+    HuggingFace Trainer's ``max_grad_norm``. ``None`` or ``0`` disables clipping."""
+
     grad_checkpointing: bool = False
     """Whether to use gradient checkpointing during the forward pass."""
 
--- a/bergson/magic/cli.py
+++ b/bergson/magic/cli.py
@@ -218,6 +218,7 @@ def worker(
         log_fn=log_fn,
         resume=resume,
         fsdp=run_cfg.fsdp,
+        max_grad_norm=run_cfg.max_grad_norm,
     )
     if getattr(run_cfg, "save_optimizer_state", False) and global_rank == 0:
         save_second_moments_as_optimizer_pt(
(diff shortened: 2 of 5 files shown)


davidoj:
Claude authored, looks reasonable to me

luciaquirke:
The trainer has historically been functional, by default I think it should continue to be functional so have the grad norm passed in per step (having stateful constant properties seems chill in general though, if this is the first stateful constant property we can probably ignore this consideration)

luciaquirke (approved):
LGTM with functional pattern + verbose comments removed

PR #304, opened 2026-06-29
Unify inversion logic

luciaquirke:
- Further unify the preconditioning with an InversionConfig
- - Add some more inversion options to match SOTA in literature, e.g. factored tikhonov for EK-FAC
- Support the factored preconditioners in Attributor/`bergson query`.
- Remove support for autocorrelation matrices in precisions other than FP32
- Rename some greek letters to be descriptive 

This PR is the other half of / will be merged with https://github.com/EleutherAI/bergson/pull/301


bergson/config/config.py
```diff
@@ -8,6 +8,8 @@
 import torch
 from simple_parsing import Serializable, field
 
+from ..hessians.inversion import Inversion
```
LouisYRYJ, line 11: can we remove this? Feels cleaner to just have the literal string here and not import from Inversion, but no strong opinion

bergson/config/config.py
```diff
@@ -467,6 +491,18 @@ class QueryConfig(Serializable):
...
+    preconditioning (H^(-1/2) applied to query and index) will be used
+    if ``unit_norm`` is enabled."""
+
+    ev_correction: bool = False
+    """Use the corrected eigenvalues of a factored Hessian at ``hessian_path``
+    (requires it to have been fit with ``HessianConfig.ev_correction=True``)."""
+
+    inversion_cfg: InversionConfig = field(default_factory=InversionConfig)
+    """How to invert the Hessian at ``hessian_path``."""
+
```
LouisYRYJ, line 505: seems likethis should be subsumed by InversionConfig

bergson/hessians/preconditioner.py
```diff
@@ -0,0 +1,353 @@
...
+                vector_nsa=g.transpose(-2, -1), matrix_cb=q_g
+            ).transpose(-2, -1)
+            g = self._mul._transpose_matmul(vector_nsa=g, matrix_cb=q_a)
+            self.logger.debug("%s: rotated back (H^-1 G)", name)
+
+            out[name] = g.reshape(flat.shape[0], -1).to(flat.dtype)
+        return out
+
+
+def is_factored_hessian(hessian_path: str | Path) -> bool:
```
LouisYRYJ, line 297: We should probably say kronecker_factored

bergson/hessians/inversion.py
```diff
@@ -0,0 +1,198 @@
...
+Writing ``c = damping_factor`` and ``λ`` for an eigenvalue, with ``mean(λ)`` the
+mean over the whole spectrum:
+
+- ``damped_inverse``: ``1 / (λ + c·mean(λ))`` — uniform Tikhonov damping.
+- ``cauchy``: ``λ / (λ² + α²)`` with ``α = c·mean(λ)`` (Lorentzian filter).
+- ``pseudoinverse``: ``1/λ`` where ``λ > c·mean(λ)``, else ``0``.
+- ``factored_tikhonov``: damping split across the Kronecker activation (A) and
+  gradient (G) factors (see :func:`factored_tikhonov_damped`). This has no dense
+  analogue (there is no A/G split for a dense Gram), so on the dense path it
+  falls back to ``damped_inverse``.
```
LouisYRYJ, line 22: Why is this called cauchy? I can't find a source on this

bergson/hessians/inversion.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def damped_inverse_eigfn(
+    eigvals: Tensor, mean: Tensor, damping_factor: float
+) -> Tensor:
+    """Uniform Tikhonov damped inverse
+    ``Let λ = eigenvals, c = damping_factor.``
+    ``1 / (λ + c·mean(λ))``.
+    """
+    return (eigvals + damping_factor * mean).reciprocal()
```
LouisYRYJ, line 54: in here and the following I would make sure to say that the eigvals may be sharded. To avoid the temptation of removing the mean as argument and compute it in the function as mean(eigenvals)

bergson/hessians/inversion.py
```diff
@@ -0,0 +1,198 @@
...
+    gradient (G) factors via the Martens & Grosse trace ratio
+    ``π = sqrt(mean(λ_A) / mean(λ_G))``::
+
+        λ + π·√(c·mean(λ))·λ_G + (√(c·mean(λ))/π)·λ_A + c·mean(λ)
+
+    ``eigvals`` is the EKFAC eigenvalue grid ``λ`` (= ``λ_G ⊗ λ_A``); ``factor_a``
+    / ``factor_g`` are the per-factor eigenvalues ``λ_A`` ``[I]`` / ``λ_G`` ``[O]``.
+    All means are passed in so the distributed path can supply globally-reduced
+    values rather than per-shard ones. Reciprocate the result for the inverse.
+    """
```
LouisYRYJ, line 123: hmmh I think this will only make sense when we dont do EV correction, are we accounting for that?

bergson/hessians/inversion.py
```diff
@@ -0,0 +1,198 @@
...
+    else:
+        multiplier = EIGENFNS[inversion](eigvals, mean, damping_factor)
+
+    frac = -power
+    if frac != 1.0:
+        multiplier = multiplier.clamp_min(0).pow(frac)
+    return multiplier
+
+
+def invert_psd_matrix(
```
LouisYRYJ, line 179: this seems like only relevant for the DensePreconditioner? Then we should make it a method of that imo

bergson/hessians/preconditioner.py
```diff
@@ -0,0 +1,353 @@
...
+        """Single-process: load the full factors (concatenating any shards)."""
+
+        def load(sub):
+            return _to(_load_full(hessian_path, sub), device)
+
+        def load_replicated(sub):
+            # factor_eig_a is the full [I] in every shard, not row-sharded, so
+            # take a single shard rather than concatenating across shards.
+            return _to(_load_shard(hessian_path, sub, 0), device)
+
```
LouisYRYJ, line 158: Hmmh I am very confused by this. I don't understand where the asymmetry comes from of doing things along the [I] but not [O] dimension

bergson/hessians/preconditioner.py
```diff
@@ -0,0 +1,353 @@
...
+    )
+    if not shards:
+        raise FileNotFoundError(f"No shard_*.safetensors found in {shard_dir}")
+
+    loaded = [load_file(s) for s in shards]
+    keys = loaded[0].keys()
+    return {name: torch.cat([shard[name] for shard in loaded], dim=0) for name in keys}
+
+
+def _load_shard(hessian_path: str | Path, subdir: str, rank: int) -> dict[str, Tensor]:
```
LouisYRYJ: I am unsure how important that is, but we should ideally specify in the load_file function itself which device to load it into. This should save us moving things between cpu and gpu (or one gpu to another) and then we don't need the _to function anymore?

bergson/hessians/preconditioner.py
```diff
@@ -0,0 +1,353 @@
...
+
+        self.eigen_a = eigen_a
+        self.eigen_g = eigen_g
+        self.lambdas = lambdas
+        self.factor_eig_a = factor_eig_a or {}
+        self.factor_eig_g = factor_eig_g or {}
+        self.inversion_cfg = inversion_cfg or InversionConfig()
+        self.apply_fn = apply_fn
+        self.power = power
+        self._mul = ShardedMul()
```
LouisYRYJ: We should call self._shard_computer or sth like that, and maybe also rename the class fo ShardedComputation. It does multiple things, some which are not mults and also it does different kind of mults, so changing the name makes sense probably

bergson/hessians/apply_hessian.py
```diff
@@ -91,69 +96,38 @@ def compute_ivhp_sharded(self):
...
+        # host read, so the sync must sit between the copies and it; the earlier
+        # bug synchronized *before* issuing the copies, leaving NaNs.
+        staged = {
+            name: v.to(device="cpu", non_blocking=True)
+            for name, v in transformed.items()
+        }
+        if torch.cuda.is_available():
+            torch.cuda.synchronize()
+        for name, t in staged.items():
+            grad_buffer[name][:] = t.flatten(1).numpy()
```
LouisYRYJ, line 130: is this the same bug we had ran into at a much earlier point? Great find!

bergson/hessians/sharded_computation.py
```diff
@@ -100,58 +109,52 @@ def _transpose_matmul(
...
 
-            dist.broadcast(shard_ci, src=rank_index)
-
-            matrix_noi[:, start_row:end_row, :].mul_(fn(shard_ci))
+            dist.broadcast(shard, src=rank_index)
+            matrix_noi[:, start_row:end_row, :].mul_(shard)
 
             if self.rank != rank_index:
-                del shard_ci
+                del shard
```
LouisYRYJ, line 169: can we put this into a _sharded_scale_rows_in_place method like with the other functions? this will also help keep track of shapes and what we compute and how

PR #313, opened 2026-07-08
Fix FAISS ANN indices crashing on CPU load

davidoj:
david reviewed, made some minor changes, and it looks reasonable

---

## Summary

Any IVF/ANN FAISS index mmap'd from disk crashed on load; only an exact `Flat`
index survived. `FaissIndex.__init__` reads each shard with `IO_FLAG_MMAP` (a CPU
index) and, when `mmap_index=False` (the shipped default), passed it to
`index_to_device(shard, "cpu")`. That helper's CPU branch calls
`faiss.index_gpu_to_cpu`, which clones the index and raises
`RuntimeError: clone not supported for this type of inverted lists
OnDiskInvertedLists` for any mmap'd IVF/ANN index.

## Changes

- `bergson/query/faiss_index.py`: `index_to_device` is now self-guarding — it
  detects whether the index actually lives on a GPU (`_is_gpu_resident`: a direct
  `GpuIndex`, or the `IndexShards` / `IndexReplicas` container that a multi-GPU
  move returns) and treats a CPU->CPU request as a no-op, returning the index
  unchanged instead of cloning via `index_gpu_to_cpu`. Both call sites
  (`FaissIndex.__init__` and `create_index`) just call the helper without device
  guards. GPU build/query behaviour is unchanged.
- `bergson/config/config.py` + `bergson/query/query_index.py`: the `query` CLI
  previously hardcoded `FaissConfig()`, so ANN / `mmap_index=True` were
  unreachable. `QueryConfig` now carries a nested `faiss_cfg: FaissConfig`,
  exposing `--faiss_cfg.index_factory` (e.g. `IVF1024,SQfp16`),
  `--faiss_cfg.mmap_index`, `--faiss_cfg.nprobe`, `--faiss_cfg.num_shards`, etc.
  Defaults reproduce the previous behaviour (exact `Flat`, in-memory).
- `tests/test_faiss_ann_cpu_load.py`:
  - End-to-end regression tests that build tiny on-disk IVF (ANN) and `Flat`
    indices and assert the CPU load path (`mmap_index=False`) loads without
    raising and returns usable neighbours (plus the `mmap_index=True` path).
  - Unit tests for the `index_to_device` guard, both directions:
    - no-op on an in-memory CPU index, and on the mmap'd OnDisk index that used
      to crash (asserting the raw `index_gpu_to_cpu` still raises, so the
      regression case is real);
    - op when it should be — a GPU-resident `IndexShards` (what a multi-GPU move
      returns) is actually converted to a new, non-container, searchable index.

## Testing

- 6 passed (ANN + Flat + mmap end-to-end, plus the three guard unit tests).
- Note: run against a `faiss-cpu` build, so the CPU no-op path (the bug) is fully
  covered; the direct-`GpuIndex` and CPU->CUDA move paths can't be exercised
  without a GPU.

https://claude.ai/code/session_01EeaHeis3YkmXwc5emPJnXo


davidoj:
Looks reasonable to me

luciaquirke (approved):
LGTM!! Thanks so much for figuring this out

PR #360, opened 2026-07-26
feat: track MoE models with fused-parameter experts and routers #318

### LINKED ISSUE #318: Support gradient tracking for MoE models with fused-parameter experts/router
Bergson currently only tracks nn.Linear, HF Conv1D, and nn.Conv{1,2,3}d modules (see LayerAdapter.supported_modules in bergson/gradients.py:238, known limitations section of focs). Modern MoE models with fused-parameter experts/routers (bare nn.Parameters rather than nn.Linear layers — e.g. gpt-oss, Mixtral, Qwen-MoE, OLMoE in transformers 5.x) are silently skipped.

We should add gradient-collection support for these fused MoE layouts so experts and routers are attributed correctly.

Notes:

- LayerAdapter (bergson/gradients.py:237-262) is the extension point.
- Should cover GPT-OSS, Mixtral, Qwen-MoE, and OLMoE.- 
- Account for per-token expert routing/sparsity when computing per-example gradients.
- Possibly the simplest solution involves letting the user manually specify bare 2D nn.Parameters, or find the 2D nn.Parameters...

We could also add support for VAEs etc.

kohsheen1234:
Closes #318 

**Problem.** `transformers` 5.x stores MoE experts as a fused 3D `nn.Parameter` and the router as a bare 2D `nn.Parameter`. Neither is an `nn.Linear`, so `discover_targets` skipped both silently — gpt-oss-20b had 5.8% of its parameters attributed, with no warning.

**Solution.** Behind `--track_moe_experts`, each expert projection becomes its own module (`...mlp.experts.expert_3.gate_up_proj`) and the router is tracked in place: 5.8% → 97.2% on gpt-oss-20b, 3.2% → 99.7% on Mixtral-8x7B. Each expert's ragged routed tokens are padded into an `[N, L, ·]` grid, so `_compute_gradient` is unchanged. Off by default, because enabling tracking replaces the fused experts' forward and costs 3–11x on collection (`benchmarks/moe_padding_overhead.py`); when off, bergson now warns that experts are being skipped instead of staying silent.

**Done.** 32 tests passing — per-example expert and router gradients match per-sample autograd across gpt-oss, Mixtral, Qwen3-MoE, OLMoE and DeepSeek-V3. 6 CUDA bf16/fp16 tests are written but **skipped and unverified** (no GPU available to me). Not covered: `llama4`/`longcat_flash` (fused params, no decorator contract) and `--attribute_tokens`.

moe_coverage

moe_tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,13 +262,17 @@ Not all models are affected — run `bergson test_model_configuration` before en
 
 See `benchmarks/` for scripts to reproduce and generate benchmarks on your own hardware.
 
-# Known limitations
+# Mixture-of-Experts models
 
-## MoE fused-parameter experts are not attributed
+In `transformers` 5.x an MoE layer stores its experts as one fused 3D `nn.Parameter` and its router as a 2D one, so neither is an `nn.Linear` and both are skipped. On gpt-oss-20b that leaves 5.8% of the parameters attributed.
 
-Bergson only tracks `nn.Linear`, HF `Conv1D`, and `nn.Conv{1,2,3}d` modules. In modern MoE models (e.g. gpt-oss, Mixtral, Qwen-MoE, OLMoE in `transformers` 5.x), the experts and router are fused bare `nn.Parameter`s rather than `nn.Linear` layers, so they are silently skipped — only attention projections and `lm_head` are tracked (~1-2% of the model's parameters).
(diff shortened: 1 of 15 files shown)


kohsheen1234:
@luciaquirke Can you please look into this :) Would love to hear feedback!

luciaquirke:
Hello, if your code and PR description are unedited by human hand such that they read as clearly agent-generated then I'm not going to read them. Excepting tests where anything goes.

Feedback: code looks bloated

luciaquirke:
We're moving documentation from the README into the docs so the new MOE info would go in there too, in the .rst format

PR #370, opened 2026-08-02
WikiText-2 / GPT-2 SOURCE replication YAMLs

luciaquirke:
Reproduces the Bae et al. 2024 WikiText-2 / GPT-2 result (**SOURCE > EK-FAC IF**,
Fig. 6) end to end from public data, with no external ground-truth artifacts.

## What's here

The replication feature line (SOURCE Adam preconditioner + Eq-43 hybrid, MAGIC
trainer optimizer.pt / interval-save support, precomputed-subset LDS, the
train / EK-FAC / SOURCE configs), plus a final round that makes it
self-contained:

- **Dataset on the Hub.** `examples/replication/prep_dataset.py` reproduces the
  512-token chunking from public `wikitext-2-raw-v1` via the standard `run_clm`
  recipe — byte-identical to the reference (4656 train / 481 validation) — and
  the configs load `EleutherAI/bergson-wikitext-2-4656-chunks` instead of a
  machine-local path.
- **`validate` scores LDS against a bergson-built bank** via `--retrain_bank` /
  `--subsets` (renamed from `retrained_dir`), so the ground truth is regenerated
  by the same production rules rather than borrowed — an independent replication,
  not a re-scoring of shipped masks/losses.
- **SOURCE auto-exports trainer checkpoints.** The pipeline converts raw
  `checkpoints/step_<n>.ckpt` DCP dirs to the HF format `from_pretrained` needs
  on demand (reusing existing exports), dropping the manual export step.
- **`examples/replication/README.md`** documents the ordered chain
  (train → EK-FAC → SOURCE → bank → validate); the machine-specific run drivers
  under `experiments/` are dropped in favour of it.

## Status

Train / EK-FAC / SOURCE scores reproduce. The ground-truth bank build and the
final LDS numbers are running; I'll post them here when they land.

PR #381, opened 2026-08-02
WikiText-2 / GPT-2 replication configs

luciaquirke:
The full replication chain against merged main: train (interval checkpoints + optimizer state) → SOURCE (auto-exported checkpoints, Adam preconditioner, `inversion_cfg.damping_factor: 1e-8`) → EK-FAC → 100 models retrained on random halves of the training data → SOURCE LDS. Dataset is the hosted `EleutherAI/bergson-wikitext-2-4656-chunks` with the script that reproduces it bit-for-bit from public WikiText-2. All configs parse against current `main`; replaces #370.

The retrain + validate steps are running now on this machine; LDS numbers will be posted here when they land.

### DIFF
--- a/examples/replicate_bae_approx_unrolling_source/README.md
+++ b/examples/replicate_bae_approx_unrolling_source/README.md
@@ -1,0 +1,17 @@
+# WikiText-2 / GPT-2 replication (Bae et al. 2024)
+
+Reproduces SOURCE > EK-FAC IF (Figure 6) end to end from public data.
+
+Run each config with `PYTHONPATH=$PWD python -m bergson <config>`, in order:
+
+| # | Config | Produces |
+|---|--------|----------|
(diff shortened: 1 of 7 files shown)


luciaquirke:
LDS (100 models retrained on random halves, α = 0.5, 481 validation queries):

| Method | LDS (mean Spearman ± 95% CI) |
|---|---|
| EK-FAC IF | 0.4182 ± 0.0155 |
| SOURCE | **0.4292 ± 0.0153** |

Absolute values sit slightly below the kronfluence-ground-truth chain (0.4451/0.4506). I think the difference is due to the independent ground-truth draw/1 model training seed, because we replicated exactly using their draw/subsets file in the working branch (see closed ADAM SOURCE PR). Ordering and magnitudes match.

PR #421, opened 2026-08-11
feat(magic): add score_trajectory report (per-step score-vs-step plot)

davidoj:
## What

Adds `bergson score_trajectory <run>` — an opt-in report you run **after** a MAGIC backward pass to *see* how attribution-score magnitude behaves across training.

It reads a finished run's `scores/` dir and `config.yaml`, groups the scores into training steps (`batch_size` rows per step), and writes `<run_path>/score_vs_step.png`: the per-step median `log10|score|` (the score **level**) vs training step.

```bash
bergson score_trajectory runs/my-magic             # -> runs/my-magic/score_vs_step.png
bergson score_trajectory runs/my-magic --window 5  # + overlay the step-norm curve & residual
```

## Why

The docs' Metasmoothness section says the game is *"identify and manage spikes"*, and `bergson metasmoothness` gives a scalar health score — but there was no way to **look** at the per-step score behaviour. On real runs the level sweeps ~tens of decades and can "ring"; this plot makes that visible at a glance, as the visual companion to the metasmoothness number. `--window N` overlays the window-N step-normalisation curve the level would be divided by (and the residual level, which should sit near 0).

## Changes

- **`bergson/magic/score_trajectory.py`** — `plot_score_trajectory()` + a reusable `per_step_level()` (pools trailing token/query axes into the step).
- **`Score_Trajectory` CLI command** wired into `__main__`.
- **matplotlib** added as an optional **`viz`** extra (and to `dev`), imported lazily so the rest of the CLI never requires it.
- **Docs:** a "Score trajectory" section in `magic.rst` + a `cli.rst` entry.
- **Tests:** CPU-only, covering `per_step_level` (known levels, trailing-axis pooling, dead steps) and end-to-end PNG output.

## Testing

- `pytest tests/test_score_trajectory.py` — 4 passed; `tests/test_cli.py` still green (command registry intact, 16 commands).
- pre-commit (black + ruff) clean; pyright (CI-pinned 1.1.378) clean on changed files.
- CLI smoke-tested end-to-end (`--help`, default, and `--window`) on a synthetic run dir.


davidoj:
I added a plot to check the stability of the score as it's a useful health check on the attributed scores. I also added a documentation section explaining some brief results with score normalisation. I didn't add score normalisation as a feature because it seems like it's not that helpful and sometimes harmful, but it's straightforward to implement so I figure if it's in the docs people can just ask their assistant to normalise the scores if they want to try it.

luciaquirke:
This is a good experiment but 1/3 wins and the win going from .7ish to .8ish seems maybe too low to merge to the main library if it would give people false hope hmmmm

PR #427, opened 2026-08-21
Custom GPT-2 with optional QK-norm and MLP pre-activation LayerNorm

luciaquirke:
## Why

The metasmoothness paper's architecture axis needs GPT-2 variants (OLMo-style QK-norm; pre-activation norm) whose effects are separable from "effect of reimplementing GPT-2" (design ruling D10 in the experiment repo).

## What

One class, three architectures via `config.arch_mod` (`none` / `qk_norm` / `preact_layernorm`):

- Modifications attach as **forward hooks on the stock `c_attn` / `c_fc` modules** — no module replacement, no forward-method copying — so the stock state-dict layout, cache handling, and attention implementation are inherited unchanged from the installed transformers.
- `arch_mod="none"` is therefore structurally stock GPT-2: identical parameter set, and **bit-identical logits under real gpt2-124M weights** (tested).
- `qk_norm` follows the OLMo 2 convention: RMSNorm on the full q/k projections before the head split, one pair per layer.
- `preact_layernorm`: LayerNorm on the MLP pre-activation, between `c_fc` and the GELU.
- Variants select at load time — `from_pretrained(..., arch_mod="qk_norm")` — which is exactly what bergson's `model_kwargs` plumbing passes through; no bergson code changes needed to train them.
- `examples/gpt2_custom/push_to_hub.py` published the model with remote code as `EleutherAI/gpt2-custom` (private); Hub round-trip with an `arch_mod` override is verified.

## Testing

`tests/test_gpt2_custom.py`: 6 passed — none-mod state-dict + bit-identical-logit equivalence, per-mod parameter counts, gradient flow to the new norms, save/load round-trips preserving architecture and function; plus the full-size gpt2-124M bit-identity check and a Hub `trust_remote_code` load.

### DIFF
--- a/bergson/models/modeling_gpt2_custom.py
+++ b/bergson/models/modeling_gpt2_custom.py
@@ -1,0 +1,84 @@
+"""GPT-2 with optional architecture modifications, for attribution experiments.
+
+One class, three architectures selected by ``config.arch_mod``:
+
+- ``"none"``: exactly stock GPT-2. No extra modules, no extra parameters; stock
+  ``gpt2`` checkpoints load with zero missing/unexpected keys, and logits match
+  ``GPT2LMHeadModel`` bit-for-bit (tested).
+- ``"qk_norm"``: the OLMo 2 convention — RMSNorm applied to the query and key
+  projections over their full width, before the head split, one norm pair per
+  layer.
+- ``"preact_layernorm"``: LayerNorm on the MLP pre-activation (between ``c_fc``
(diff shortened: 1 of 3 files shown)


luciaquirke:
Moved to the metasmoothness experiment repo per Lucia — this is paper-specific code with no bergson dependency (the Hub repo EleutherAI/gpt2-custom carries the remote-code copy runs actually load). Branch deleted.

PR #429, opened 2026-08-21
Route query evaluation through the training micro-batcher

luciaquirke:
## Why

Query evaluation had its own forward loops, so it missed the micro-batching every training step gets and ignored the zero weights padding puts on its duplicate rows.

## What

- `mean_query_loss` replaces the three copies of the scalar query-loss loop. It and `per_doc_query_losses` split each batch with `split_batch`, the same splitter a training step uses, so `grad_accum_steps` bounds their activation memory too:

  |  | gas=1 | gas=4 | gas=8 |
  |---|---|---|---|
  | train step | 59.07 GiB | 16.26 GiB | 8.61 GiB |
  | `mean_query_loss` | 20.73 | 7.28 | 3.64 |
  | `per_doc_query_losses` | 20.63 | 5.16 | 2.58 |

- `mask_padded_rows` masks the zero-weight rows out of the query loss and the query gradient. Padding fills a batch by repeating the last document, and both paths had dropped `example_weight`, so the copies counted as real query documents: a 5-document query set at `batch_size` 64 was 92% copies of document 4. The rows are masked rather than removed so every rank runs the same number of micro-batches, which FSDP's forward all-gather requires. Both reductions divide by the all-reduced count of batches with a live row, so a batch of nothing but padding contributes nothing.

- `weighted_causal_lm_ce` divides the summed token loss by the clamped supervised-token count. Identical to `reduction="mean"` wherever anything is supervised, and 0 rather than NaN for a micro-batch of only padding.

- The per-query stream is one row per rank instead of `batch_size` rows. Any wider is zero-weight padding that still costs a full `batch x seq x vocab` slab of logits per rank (about 26 GB at batch 256 for GPT-2). The single-document dataset also drops its `doc_ids` column, because the stream indexes its weights by row.

- `mean_query_loss` all-reduces, as `per_doc_query_losses` already did, which also fixes `evaluate_retrained` reporting only the local shard's loss under DDP. `validate_scores` loses its now-unused `world_size` parameter.

GPT-2, 256 WikiText chunks, 5 query documents at `batch_size` 64: aggregate query loss 4.0460 before, 4.3320 after; scores correlate at 0.941 and differ by up to 59% of the largest score. Per-query MAGIC is unaffected: each of its batches is one document repeated.

## Tests

`test_query_eval_ignores_padding_and_grad_accum` checks that a padded query stream at `grad_accum_steps` 1 and 4 gives the same per-document losses, aggregate loss and query gradient as the unpadded stream at 1, and that masking keeps the batch width. It fails on the previous code. `test_unsupervised_batch_loss_is_zero` pins the loss of an all-padding batch.

### DIFF
--- a/bergson/config/config.py
+++ b/bergson/config/config.py
@@ -412,6 +412,14 @@ class TrainingConfig(AttributionConfig, Serializable):
     under any split; ignored under dropout (``train_mode``), where the replay
     must reuse the forward's micro-batches."""
 
+    eval_batch_size: int = 0
+    """Batch size for query-loss and query-gradient evaluation. ``0`` selects
+    ``min(batch_size, 32)``. A memory knob only: query evaluation materialises
+    fp32 logits of ``batch x seq x vocab`` (~26 GB at batch 256 for GPT-2), and
+    zero-weight padding rows contribute nothing to the weighted mean, so the
+    eval batch never affects results — training batch is the science, eval
+    batch is the memory. Rounded down to a multiple of the world size."""
+
     resume: bool = False
     """Resume a previously interrupted run from the last checkpoint."""
 
--- a/bergson/config/config.py
+++ b/bergson/config/config.py
@@ -414,11 +414,22 @@ class TrainingConfig(AttributionConfig, Serializable):
 
     eval_batch_size: int = 0
     """Batch size for query-loss and query-gradient evaluation. ``0`` selects
-    ``min(batch_size, 32)``. A memory knob only: query evaluation materialises
-    fp32 logits of ``batch x seq x vocab`` (~26 GB at batch 256 for GPT-2), and
-    zero-weight padding rows contribute nothing to the weighted mean, so the
-    eval batch never affects results — training batch is the science, eval
-    batch is the memory. Rounded down to a multiple of the world size."""
+    ``min(batch_size, 32)``; rounded down to a multiple of the world size.
+
+    A training step bounds its activation peak by splitting each batch into
+    ``grad_accum_steps`` micro-batches, but the query-loss evaluators in
+    :mod:`bergson.validate` run a whole per-rank batch in one forward and add a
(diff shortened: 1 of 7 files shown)


luciaquirke:
Escalation from the A40 fleet: this is now gating five of thirteen batch/knob rows, not just bs512 — every bs256 row OOMs at the MAGIC backward on 47.5 GB A40s (constant 1.53 GiB request against 42.75 GiB allocated + 3.58 GiB reserved-unallocated fragmentation; `expandable_segments` recovered bs64/bs128 but bs256 needs the eval cap itself). Evidence in the experiment repo, messages/2026-08-21-adam-backward-oom-a40.md.

luciaquirke:
Verified the micro-batcher rework on the A100 node: `test_per_query_magic` + `test_magic` accumulation/per-query/eval selections (15) and `test_multi_query_validate` (7) all pass. Routing through the verified-exact accumulation path is cleaner than the parallel knob I added — and it covers the validate path too. Ready from the fleet's side; five A40 rows unpark on merge.

luciaquirke:
Head `18d1e516` verified on the A100 node: 28 tests pass (per-query, multi-query validate, accumulation). The padded-rows denominator fix checks out. From the fleet's side this is merge-ready; two A40 rows are parked specifically on it and three more behind them.

luciaquirke:
Appended (no rewrite) one commit: the **per-query** stream still padded single query documents to the full training batch — the path every paper row uses (`query_method: none`). bellflower-0's diagnostic was conclusive: OOM at nproc 2, 4, *and* 8 on A40 (each rank pads the full width independently, so sharding can't help). Capped at `min(batch_size, 32)` rounded to a world-size multiple, same inert-padding argument as the aggregate fix; 21 tests pass unmodified. This is the commit that actually unblocks the five parked bs256 rows — and 32k/64k on any A40.

PR #431, opened 2026-08-22
Reduce aggregate query evaluation over tokens, not over batches

luciaquirke:
Stacked on #429 (`fix/eval-batch-size`), which is where `mean_query_loss` and `mask_padded_rows` live. Retarget to `main` once that merges.

## Why

#429 stopped padding rows from entering aggregate query evaluation. What is left is the reduction over the batches themselves: both `mean_query_loss` and `compute_query_gradients` average the *per-batch* means. Each batch reports the mean over its own label tokens, and those means are then averaged with equal weight, so a batch holding one live document counts as much as a batch holding eight. The batch width, a memory and throughput knob, lands in a recorded loss and in the query gradient the MAGIC scores are built from.

Ten GPT-2 query documents, padded to a batch width of 4:

| eval batch | pad rows | aggregate query loss | mean over the real documents |
|---|---|---|---|
| 4 | 2 | 4.716831 | 4.779631 |
| 8 | 6 | 4.638320 | 4.779621 |
| 32 | 22 | 4.779619 | 4.779618 |
| 256 | 246 | 4.779622 | 4.779621 |

The aggregate only agrees with the query set's own mean once every document fits in one batch.

## What

Weight each batch by the label tokens it actually carries and divide once at the end, in both `mean_query_loss` and `compute_query_gradients`. That is exactly what a single batch spanning the whole query set already produced, so the reduction now runs over the query set rather than over the batching. Dead batches (every row padding) contribute an exact zero to both sums, which the `weighted_causal_lm_ce` guard from #429 already provides.

After the change the same table is flat: 4.779631 / 4.779620 / 4.779619 / 4.779619 / 4.779622 across batch widths 4 through 256, matching the mean over the real documents at every width.

## This moves recorded numbers

End-to-end GPT-2 run, 10 query documents at batch 8 (so the second batch is 2 real documents against the first batch's 8), `query_method: mean`, leave-k-out over 3 subsets:

- baseline query loss 3.6654106974601746 -> 3.5614548206329344
- aggregate MAGIC scores: Pearson 0.9907, Spearman 0.9270 against the old run, largest change 41% of the largest score

That case is deliberately lopsided. The standard 50-query configs feel a milder version of the same effect rather than none: 50 documents sharded over 4 ranks live 13/13/12/12, and the old reduction weighted those four rank means equally.

`query_method="sum"` changes scale: it is now the gradient of the summed query loss rather than of the summed per-batch means. No config in the repository sets it.

## Testing

- `test_aggregate_query_reduction_ignores_batch_width`, new: the same 5 unequal-length documents as one batch, and as three batches whose last is 1 real + 1 padding, must give the same aggregate loss and the same query gradient; the loss must equal the token-weighted mean of `per_doc_query_losses`. Fails on `fix/eval-batch-size` (both parametrisations), passes here.
- `tests/test_per_query_magic.py tests/test_magic.py tests/test_bank_loss_cache.py tests/test_banked_model_loading.py tests/test_multi_query_validate.py tests/test_per_token_lds.py tests/test_batch_size_invariance.py`: 75 passed.
- `pre-commit run --all-files` and `pyright` clean.

https://claude.ai/code/session_01MpLMt4jYaPpjUUz4PoBbja

### DIFF
--- a/bergson/magic/cli.py
+++ b/bergson/magic/cli.py
@@ -48,7 +48,7 @@ from ..utils.worker_utils import setup_data_pipeline
 from ..validate import validate_scores
 from .config import MagicConfig
 from .data_stream import DataStream, mask_padded_rows, pad_dataset_to_batch_size
-from .grad_accum import accumulate_grads
+from .grad_accum import accumulate_grads, loss_denom
 from .trainer import BackwardState, TrainerState, prepare_trainer, write_lr_history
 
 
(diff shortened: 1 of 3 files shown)


luciaquirke:
Superseded by #432, which lands the same token-weighted reduction with two differences found by running it on multiple GPUs: the weight scales the loss (via `accumulate_grads(loss_scale=...)`) rather than the gradient, because FSDP sums gradients across ranks inside autograd and scaling afterwards failed 4/5 of `test_distributed_magic.py`; and `query_method: none` still divides, so the reported per-query baseline stays a mean. #432 also adds a 2-GPU test comparing every rank to single-process.

PR #436, opened 2026-08-22
Compare a tail filter against random filters of the same size

luciaquirke:
Stacked on #430 — review that first; this branch contains its commits.

A `filter-proponents` loss change means nothing on its own: removing any 1% of
the training data moves the query loss. Without a reference the number is
unreadable, and every use of it so far has compared it against a random filter
run by hand.

The filter arm now runs alongside `num_subsets` random removals of the same
size, and reports the filter's rank among them per query:

```
filter-proponents: mean loss change 0.001111 over 3 queries (10 of 200 docs removed per query, 5.000%)
Random baseline (4 subsets of 10 docs, retrained here): mean loss change 0.003895
Query 0: filter-proponents -0.010040  random -0.000756 +/- 0.014530  rank 4/5
Query 1: filter-proponents +0.002502  random +0.004464 +/- 0.015796  rank 4/5
Query 2: filter-proponents +0.010871  random +0.007976 +/- 0.013704  rank 4/5
```

Rank 1 is the largest loss increase. With four or five draws a rank is what the
sample supports; the mean and sd are there too, but no p-value is claimed.
`random_filter.csv` holds the per-draw losses and `filter_summary.csv` the
comparison.

### Reusing a bank instead of retraining the arm

The random arm ignores the scores, so one retrain serves every query — and an
LDS bank already holds exactly those retrains. `retrained_dir` therefore
supplies the random arm for a `filter-*` run, where it previously shortcut
`lds` alone. No new flag: the same "read it from the bank" idiom, applied to
the one arm a bank can serve. The filter arm is score-dependent and always
retrains.

`num_subsets: 0` with no bank skips the arm.

A bank drawn at a different removal size, `subset_weight`, or
`exclude_zero_scores` pool is not a comparable baseline, so a mismatch raises
rather than reporting an apples-to-oranges number. With `subset_fraction: 0`
the filter's size is `1/num_subsets` of the pool — the LDS chunk size — so a
default bank lines up as it is. `subsets:` reuses an LDS run's draws when the
models were not saved.

The bank's per-subset query losses come from the same cache
`evaluate_retrained` fills, lifted out as `load_bank_losses` so both callers
share it.

### Testing

Unit tests for the draws, the bank rejections and the ranking; two end-to-end
runs on a tiny model, one asserting the arm's outputs and one asserting that
reading the arm from a bank reproduces retraining the same subsets locally. On
a 200-doc gpt2 run the two agree to 1e-6.

`tests/test_validate_routing.py` also picks up the behaviour-based rewrite that
missed the #435 merge — the previous tests asserted the kwargs `Validate`
passes to `run_magic`, which broke on the new argument and asserted nothing
about what the commands do.

### DIFF
--- a/bergson/cli/commands.py
+++ b/bergson/cli/commands.py
@@ -306,18 +306,26 @@ class Validate(ValidationConfig):
     baseline_model: str = ""
     """Optional path to baseline model trained on the full dataset."""
 
+    @property
+    def retrained_dirs(self) -> list[str]:
+        """``retrained_dir`` normalized to a list of paths."""
+        if isinstance(self.retrained_dir, str):
+            return [d for d in self.retrained_dir.split(",") if d]
+        return list(self.retrained_dir)
+
     def execute(self):
         """Run the validation."""
         assert self.scores, "Path to attribution scores must be provided."
-        if self.retrained_dir:
-            retrained_dir = self.retrained_dir
-            if isinstance(retrained_dir, str) and "," in retrained_dir:
-                retrained_dir = retrained_dir.split(",")
-            evaluate_retrained(self, retrained_dir, score_path=self.scores)
+        dirs = self.retrained_dirs
+        # A filter-* run still retrains its score-dependent arm, so a bank
+        # replaces the whole run only for lds.
+        if dirs and self.method == "lds":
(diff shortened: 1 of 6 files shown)


luciaquirke:
Rolled into #430.

PR #440, opened 2026-08-24
EK-FAC: average query gradients over model checkpoints

luciaquirke:
- Add `HessianPipelineConfig.query_model_paths` for averaging a query's gradients for multiple checkpoints.
- With two or more paths, the query gradient is built once per checkpoint and averaged before the Hessian is applied.

### Implementation

- `average_gradient_indices` in `bergson/data.py` averages several gradient indices into a new one. Accumulates in float64 for safety.
- `hessian_pipeline` builds each checkpoint's query index into `<query_path>__ckpt<i>`, then averages them into `query_path`.

### DIFF
--- a/bergson/config/config.py
+++ b/bergson/config/config.py
@@ -906,17 +906,11 @@ class HessianPipelineConfig:
     """Config for the Hessian-preconditioned influence pipeline."""
 
     query_model_paths: list[str] = field(default_factory=list)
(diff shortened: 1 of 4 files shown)


luciaquirke:
cc @LouisYRYJ I think I merged the MAGIC one already but this one looks a bit more messy

PR #456, opened 2026-09-09
Add TRAK

luciaquirke:
- Add `bergson trak` for [TRAK (Park et al., 2023)](https://arxiv.org/abs/2303.14186).
- `trak_cfg.checkpoints` averages the scores of independently trained models. 
- Refuse incompatible projection/loss function settings
- Stream the global projection per-module to avoid OOM, because the paper uses a 4k+ projection dim. This is mathematically different from the per-module projection of TrackStar.
- New `loss_fn: margin`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ To build a train‑time gradient store, use our HF Trainer callback. This will i
 
 Bergson supports on-disk gradient stores and on-the-fly queries, and per-token and per-sequence attribution.
 
-`bergson trackstar` and `bergson ekfac` both orchestrate multi-step attribution recipes over a model checkpoint. `bergson trak` runs [TRAK](https://arxiv.org/abs/2303.14186): projected gradients whitened by their damped Gram inverse, weighted by one minus the training example's label probability, optionally ensembled over checkpoints.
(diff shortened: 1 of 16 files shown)


luciaquirke:
Follow-up commit: the first cut used Bergson's per-module `autocorrelation` Hessian, i.e. a block-diagonal approximation of TRAK's kernel. Now `trak` defaults to `kernel: joint`: a new `gram` Hessian method concatenates every module's projected gradient (index order) and fits one Gram over the concatenation; `JointDensePreconditioner` applies its damped inverse as a single matrix. `kernel: per_module` keeps the block-diagonal variant. `projection_target: global` (a single global sketch) also works with the joint Gram, but Bergson materialises its k×d projection blocks, so for GPT-2-size models the per-module double-sided sketch is the practical choice (48 modules × 16² = 12,288 dims).

Tests: `test_trak_joint_kernel_matches_explicit_formula` builds train/query indices with `build`, forms Φ, and checks the pipeline's scores against φ_qᵀ(ΦᵀΦ/N + λ)⁻¹φ_i (agreement to the index's storage precision); the trackstar-equivalence and ensemble tests now run with `kernel: per_module`. All pass on an A40 along with `test_global_projection.py` and `test_hessian.py`.

luciaquirke:
Renamed per review: the joint variant is now `HessianConfig.scope: per_module | joint` on the single `autocorrelation` method (same matrix, different scope), collector `JointAutocorrelationCollector`; `TrakConfig.kernel` sets the scope. Tests re-run on an A40 (trak + hessian suites pass).

Issue #453, opened 2026-09-03
Checkpoint Hessian fit


luciaquirke:
The KFAC fit is a large accumulation but doesn't save progress checkpoints.

Every N documents, rank 0 should write the accumulator tensors + doc cursor to `<run_path>.part/fit_state.pt`; load it in on resume  and skip the stream to the cursor. The same thing should happen in the correction pass.

prajak002:
Hi, I'm Prajak, an ML engineer (background in ASR, 3D reconstruction, medical imaging, and some causal inference / interpretability adjacent work). New to bergson and to EleutherAI's community. This issue looked like a good first real contribution here, so I went ahead and opened a PR for it: #470

Quick summary of the approach: since each rank only holds its own row shard of the KFAC accumulators, I'm checkpointing per rank (`fit_state_shard_{rank}.pt`) instead of only from rank 0, along with the batch cursor and running token count. It writes every `checkpoint_interval` batches, a new opt-in `HessianConfig` field that's off by default. Resume happens automatically on restart if a checkpoint file is found.

One thing I left out on purpose: checkpointing the eigenvalue correction pass, which I know the issue mentions too. Happy to add that here or as a follow up, whichever you'd prefer. Let me know if the per-rank-shard approach makes sense or if you had something else in mind.

PR #470, opened 2026-09-16
Checkpoint the KFAC fit for resumable Hessian fitting

prajak002:
Addresses #453.

## What

The KFAC covariance fit is a big accumulation over the whole training set, and right now there's no way to recover progress if it gets interrupted. This adds opt-in periodic checkpointing to the fit loop:

- Every `checkpoint_interval` processed batches (new `HessianConfig` field, default `0` so it's off by default), each rank writes its own accumulator shard to `<run_path>.part/fit_state_shard_{rank}.pt`. That file holds `A_cov_dict`, `S_cov_dict`, the batch cursor, and the running collected token count. It writes to a temp file first and renames it, so a crash mid write can't leave behind a corrupt checkpoint.
- On restart, if a checkpoint exists it gets loaded automatically and already processed batches are skipped, picking up right from the saved cursor.

## Why per-rank files instead of a single rank 0 file

Each rank only ever holds its own row shard of `A_cov_dict`/`S_cov_dict` (you can see this already in how `teardown()` saves per-rank `shard_{rank}.safetensors` files), so a single rank 0 checkpoint would quietly drop every other rank's shard. Batches are processed in lockstep across ranks anyway, since the forward/backward hooks do a blocking all-reduce every step, so the cursor stays consistent across ranks even though the accumulator content is different on each one.

## Design

`save_checkpoint`/`load_checkpoint` are added as no-op hooks on `HookCollectorBase`, along with a `checkpoint_interval` field, so the other collector types (`tkfac`, `shampoo`, `autocorrelation`, `LambdaCollector`) don't need any changes, they just inherit the no-op. `CollectorComputer.run_with_collector_hooks` calls `load_checkpoint()` once at the start and `save_checkpoint()` after every batch.

Only `CovarianceCollector` (kfac.py) actually implements them for now.

## Known gap

The issue also mentions checkpointing the eigenvalue correction pass (`ev_correction=True`, `LambdaCollector`). I kept this PR scoped to the main KFAC fit accumulation since that's the expensive part the issue is really about, and left the correction pass for a follow up rather than rush something I couldn't fully verify. Happy to add it here or separately, whatever you'd prefer.

## Testing

- `uv run pytest tests/test_hessian.py` passes.
- Wrote a standalone script that exercises `save_checkpoint`/`load_checkpoint` directly against a tiny model: checkpoint write/skip cadence, exact round trip of accumulator values, cursor, and token count, plus the missing-file and disabled-interval cases. All pass.
- I don't have a GPU in this environment, so I couldn't run a full distributed multi-rank fit end to end locally. `tests/ekfac_tests/` (the heavier integration suite) was still running CPU only after 15+ minutes when I opened this PR. Flagging that honestly instead of claiming full coverage, happy to iterate based on CI or review feedback.

I'm new to the project, would appreciate a look, thanks!


luciaquirke:
Alright this is beautiful, thank you! There are a few non-obvious details to address before merge:

- In the CLI we use the flags `--resume` and `--overwrite` respectively so users can't overwrite valuable data by mistake. The Hessian resume should only happen when the user passes --resume. 
  - The `bergson ekfac --resume` CLI command currently uses the function `prepare_step` to clean up partial steps. In other words, it currently handles resumes at a per-step granularity, whereas with this pipeline we're introducing a more granular resume _within_ a particular step of the pipeline. We need some new logic in `prepare_step` to handle this.
- We would like to support resuming all the kronecker factored Hessian fits that are accessible from the `bergson ekfac` CLI tool, including ek-fac and shampoo as well as kfac. They use different collectors, like ShampooCollector and LambdaCollector, that would also need your new method copied in. 

My coding agent also correctly raised this race condition:

> In multi-rank runs, each rank saves independently with no barrier. If a crash occurs between rank 0's rename and rank 3's, the ranks reload different cursors, iterate different numbers of remaining batches, and the blocking all-reduces in the hooks deadlock. Fix: A barrier before the rename, plus an all-gather of the cursor on load that errors on disagreement.

Let me know if you need clarification on that one.

luciaquirke:
Maybe we could have a keyword like

```
 def prepare_step(path, *, resume: bool, resumable: bool = False) -> bool:
      if resume and step_state(path) == "complete":
          return False

      part = partial_path(path)
      if part.exists() and not (resume and resumable):
          shutil.rmtree(part)
      superseded = _superseded_path(path)
      if superseded.exists():
          shutil.rmtree(superseded)
      return True
```

Not sure

luciaquirke:
I'll run the GPU tests on my machine and make any necessary fixes when this is ready, if that works for you!

prajak002:
Good catch on the resumable flag  ... i checked it against the actual flow and it's not optional, it's a real bug fix.. u can see here approximate_hessians writes the checkpoint shards into .part, but lso check hessians/pipelines.py  which have 100 calls prepare_step on that same path before the fit runs, and today it unconditionally rmtrees .part regardless of resume. So as it stands, resuming would wipe the checkpoint before "load_checkpoint()" ever sees it  .. your suggestion is exactly what makes the new checkpointing actually reachable on resume.

prajak002:
for your reference
Decide whether to run the step writing to `path`, clearing stale output.
 An interrupted `.part` is normally removed here so the step restarts from
  scratch. Pass `resumable=True` for steps that can pick up from their own
    partial output (e.g. a checkpointed fit) -- with `resume=True` that leaves
    `.part` in place instead of deleting it.
    
    
def prepare_step(path: Path | str, *, resume: bool, resumable: bool = False) -> bool:
  
    path = Path(path)

    if resume and step_state(path) == "complete":
        return False

    part = partial_path(path)
    if part.exists() and not (resume and resumable):
        shutil.rmtree(part)
    superseded = _superseded_path(path)
    if superseded.exists():
        shutil.rmtree(superseded)
    return True

# bergson/hessians/pipeline.py
def _step_complete(path: str, resume: bool, resumable: bool = False) -> bool:
 
    if prepare_step(path, resume=resume, resumable=resumable):
        return False
    print(f"  Skipping (already complete at {path})")
    return True

# bergson/hessians/pipeline.py, Step 2 -- the only call site that needs it
if not _step_complete(f"{hessian_path}/{method}", resume, resumable=True):

Every other _step_complete/prepare_step call site keeps the default resumable=False, so nothing else changes behavior. (Checked cli/trak.py and cli/trackstar.py too  their _step_complete is a separate, older helper that doesn't call prepare_step at all, so it's untouched.)

For the other two points:
- I'll copy save_checkpoint/load_checkpoint into the shampoo and lambda (ek-fac correction) collectors so .. resume covers all three methods reachable from bergson ekfac, not just kfac.
- Agreed on the race condition  a crash between ranks' renames could leave them loading different cursors and deadlocking on the hooks' all-reduce. I'll add a barrier before the rename plus an all-gather-and-compare on the loaded cursor that errors out on disagreement, rather than silently diverging.

Will ping you when it's ready for the GPU run.... thanks for digging into this so thoroughly.

PR #472, opened 2026-09-17
Remove a dead CI workflow

prajak002:
This workflow file is entirely commented out, including the `on:` trigger block, so it reports `No event triggers defined in on` as a failure on every single push to every branch, including main. Since it isn't running anything as written, removing it rather than leaving a permanent failing check for everyone. Happy to restore a working version if it was meant to be re-enabled properly.


luciaquirke:
Thank you!! LGTM 🚀

prajak002:
Thanks Lucia!

PR #473, opened 2026-09-17
Fix test_trak.py after the QuerySetConfig migration

prajak002:
## Summary
- `_query_cfg` in `tests/test_trak.py` still built a bare `DataConfig`, but `TrakConfig.query` became a `QuerySetConfig` in #465. Every GPU-only test in this file errored on `trak_cfg.query.path` since `DataConfig` has no `path` field. Fixed by wrapping it in `QuerySetConfig(data=...)`.
- `test_trak_weights_rows_by_one_minus_p` recomputed label probabilities with a hardcoded `batch_size=4`, while the pipeline that produced the saved weights uses the default `loss_batch_size=32`. Different batching changes GPU floating-point rounding enough to fail the default `allclose` tolerance even though both values are correct. Fixed by matching `TrakConfig().loss_batch_size`.

These tests are GPU-gated (`skipif(not torch.cuda.is_available())`) and CI only runs on `ubuntu-latest`, so they never actually ran until now. Verified the fix on two different GPUs (RTX 3090, RTX 4060 Ti).

@luciaquirke flagging since this traces back to #465, in case there's other test coverage from that change that hasn't run on GPU either.

## Test plan
- [x] `uv run pytest tests/test_trak.py -v` passes on GPU
- [x] `pre-commit run --files tests/test_trak.py`


prajak002:
Hi @luciaquirke, I was testing some GPU changes and ran into an issue in tests/test_trak.py. Looks like it never got updated after the query config migration in #465, so a few tests in that file were broken. Since they're GPU only and CI runs on ubuntu-latest, nobody caught it until now. Fixed it up in this PR, figured you'd want to know since it ties back to that change. Let me know if you want me to check for anything else from #465 that might have the same problem.

luciaquirke:
This looks great, thank you so much!!

prajak002:
you are welcome !

